PR #1, opened 2017-09-25
Allow customization by experiment

ghallsimpsons:
Adds a core.Experiments enum, currently containing SPT, PB and BK.
Removes jiffybase argument from ARCFileReader. Instead, the jiffybase is set correctly based on experiment (**Note**: not yet tested with PB arcfiles).
Minor changes to arcfile parsing logic to allow reading of BK arcfiles.


nwhitehorn:
Do you see any use of this aside from processing ARC file dialects? If not, I would prefer to move the enum into the gcp library rather than core. I will look at the code soon.

ghallsimpsons:
Among the public modules, no, I can't think of any other application. I will go ahead and move the enum definition.

core/include/core/Experiments.h
```diff
@@ -0,0 +1,11 @@
+#ifndef _EXPERIMENTS_H
+#define _EXPERIMENTS_H
+
+namespace Experiments {
+        const unsigned SPT = 0;
+        const unsigned PB = 1;
+        const unsigned BK = 2;
+}
```
nwhitehorn: If it is only used in one place, you could probably make do with this being a string as well. It at least does not need to be in a header file.

gcp/src/ARCFileReader.cxx
```diff
@@ -108,13 +104,23 @@ class ARCFileReader : public G3Module {
 	std::deque<std::string> filename_;
 	std::string cur_file_;
 
+	int experiment;
```
nwhitehorn: There seems to be a type mismatch here: int in the member variable, unsigned in the constructor, and enum for the definition.

gcp/src/ARCFileReader.cxx
```diff
@@ -124,24 +130,31 @@ ARCFileReader::ARCFileReader(const std::string &path, double jiffie_base) :
 
 
 
-ARCFileReader::ARCFileReader(const std::vector<std::string> &filename, double jiffie_base) :
-    ms_jiffie_base_(jiffie_base)
+ARCFileReader::ARCFileReader(const std::vector<std::string> &filename,
+                             unsigned experiment) : experiment(experiment)
```
nwhitehorn: Please maintain standard white space (4-space line continuations) here and above.

gcp/src/ARCFileReader.cxx
```diff
@@ -124,24 +130,31 @@ ARCFileReader::ARCFileReader(const std::string &path, double jiffie_base) :
...
-		    !boost::filesystem::is_regular_file(fpath)) 
+		    !boost::filesystem::is_regular_file(fpath))
 			log_fatal("Could not find file %s", i->c_str());
 		filename_.push_back(*i);
 	}
 
 	const std::string path = filename_.front();
 	filename_.pop_front();
 	StartFile(path);
-
```
nwhitehorn: Could you strip the extraneous whitespace-only changes from the diff (here, above, and various other places)?

gcp/src/ARCFileReader.cxx
```diff
@@ -324,27 +343,38 @@ void ARCFileReader::ParseArrayMap(uint8_t *buf, size_t size)
...
 				block_info.addr = ntohl(ltmp[3]);
 				block_info.dim[0] = ntohl(ltmp[4]);
 				block_info.dim[1] = ntohl(ltmp[5]);
-				block_info.dim[2] = ntohl(ltmp[6]);
+				if ((experiment == Experiments::SPT) ||
+					(experiment == Experiments::PB)) {
+					block_info.dim[2] = ntohl(ltmp[6]);
+				} else {
+					block_info.dim[2] = 0;
+				}
```
nwhitehorn: Should the sense of this if block be inverted (BK being the odd man out as at line 348)?

gcp/src/ARCFileReader.cxx
```diff
@@ -368,31 +398,51 @@ void ARCFileReader::ParseArrayMap(uint8_t *buf, size_t size)
...
+					/* Human-readable type info */
+					memcpy(&namelen, buf + off, sizeof(namelen));
+					namelen = ntohs(namelen); off += 2;
+					memcpy(axes_help, buf + off, namelen);
+					axes_help[namelen] = '\0';
+					off += namelen;
+					block_info.axes_help = axes_help;
+				} else {
+					block_info.axes_help = "";
+				}
```
nwhitehorn: Same comment about if ()/else() ordering.

PR #8, opened 2019-05-02
Add a warning message to help debug a certain boost issue.

mhasself:
The failure was observed in a conda installation; __name__ would not
extract to string.  This built ok but import failed with uninformative
SystemError.

### DIFF
--- a/core/include/core/std_map_indexing_suite.hpp
+++ b/core/include/core/std_map_indexing_suite.hpp
@@ -466,7 +466,12 @@ return incref(tuple.attr("__iter__")().ptr());
             std::string cl_name;
             object class_name(cl.attr("__name__"));
             extract<std::string> class_name_extractor(class_name);
-            cl_name = class_name_extractor();
+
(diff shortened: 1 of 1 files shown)


core/include/core/std_map_indexing_suite.hpp
```diff
@@ -466,7 +466,12 @@ return incref(tuple.attr("__iter__")().ptr());
             std::string cl_name;
             object class_name(cl.attr("__name__"));
             extract<std::string> class_name_extractor(class_name);
-            cl_name = class_name_extractor();
+
+	    if (!class_name_extractor.check())
+	        std::cerr << __FILE__ << ":" << __LINE__ << ": object name "
+		    "extractor failed; import error imminent.\n" << std::flush;
```
nwhitehorn: Can you use the log_fatal() macro here?

mhasself: Yes, that seems to work fine.

PR #12, opened 2019-07-11
Merge MPI support

nwhitehorn:
I wrote some basic MPI support (see the README) that does frame-wise parallelization and stream joining, as well as a simple interface to TOAST built on top of that. The TOAST bit is extremely experimental (as noted in docs) but the rest is in good shape and I think is ready to merge.

One small nit is that the code relies on sending frames through MPI messages, which are limited to 2 GB. I think no more will ever make a > 2 GB frame (especially because raw timestreams are generally compressed), but it is worth thinking about possible solutions a la BigMPI.


mpi/MPIFileIO.py
```diff
@@ -0,0 +1,106 @@
...
+        self.broadcast = False
+    def nextreader(self):
+        nextfilelist = self.files.pop(0)
+        self.broadcast = False
+        if isinstance(nextfilelist, str):
+            if self.comm.rank == 0:
+                self.broadcast = True
+                self.cur_reader = core.G3Reader(nextfilelist)
+            else:
+                self.cur_reader = lambda fr: self.comm.bcast(None,root=0)
```
arahlin: This is confusing.  I'd use a single `comm.bcast` call below, set `broadcast = True` on all nodes, and just have this `cur_reader` function return `None` on all but the root node.

mpi/MPIFileIO.py
```diff
@@ -0,0 +1,106 @@
...
+
+    Rules to make this work:
+    - First module on IO nodes must be MPIFileReader
+    - Second module on IO nodes must be DeduplicateMetadata
+    - Third module on IO nodes, first on workers, must be MPIFrameParallelizer
+
+    You may want to use the MPIIODistributor pipe segment to make sure these
+    rules are followed.
+    '''
+    def __init__(self, iocomm, cpucomm, cpucomm_startrank, dataframetype=[core.G3FrameType.Timepoint, core.G3FrameType.Scan]):
```
arahlin: Need a docstring for this `__init__` method

mpi/MPIFileIO.py
```diff
@@ -0,0 +1,106 @@
...
+                # Send it somewhere random
+                self.cpucomm.send(frame, random.randint(self.cpucomm_startrank, self.cpucomm.size-1))
+
+        if frame.type == core.G3FrameType.EndProcessing:
+            return frame
+
+        return [] # Terminate processing on IO nodes
+ 
+@core.pipesegment
+def MPIIODistributor(pipe, mpicomm=MPI.COMM_WORLD, n_io=10, files=[]):
```
arahlin: Docstring

mpi/MPIFileIO.py
```diff
@@ -0,0 +1,106 @@
...
+        self.iocomm = iocomm
+        self.cpucomm = cpucomm
+        self.cpucomm_startrank = cpucomm_startrank
+        self.dataframes = dataframetype
+    def __call__(self, frame):
+        if frame is None:
+            # Initial frame is None on CPU nodes
+            frame = self.cpucomm.recv()
+            if frame.type == core.G3FrameType.EndProcessing:
+                frame = [] # This is the start module
```
arahlin: Better documentation of what is happening here.

mpi/MPIFileIO.py
```diff
@@ -0,0 +1,106 @@
...
+            return frame
+        if frame.type not in self.dataframes:
+            # Synchronize and broadcast metadata; should be the same
+            # in all streams
+            self.iocomm.barrier()
```
arahlin: barrier isn't necessary here since the bcast will create one

mpi/README.rst
```diff
@@ -0,0 +1,28 @@
...
+
+Parallelization in this sense can be achieved by replacement of G3Reader in a normal G3Pipeline with MPIIODistributor.
+
+Frame Accumulation
+==================
+
+At the end of a sequence of parallel pipelines, you in general want to join the data from all the processes in the communicator back together again. Typical cases are stitching processed scans back together again, for example to feed to a maximum-likelihood map-maker or some other algorithm that needs very large quantities of distributed again, or doing a parallel reduction, for example by coadding map frames produced by N parallel map-makers.
```
arahlin: `distributed again` -> `distributed data`

mpi/TOASTFiller.py
```diff
@@ -0,0 +1,98 @@
...
+            self.invkeys = {k: i for i,k in enumerate(self.keys)}
+
+        if frame.type == core.G3FrameType.Calibration:
+            #self.detquat = {k: qa.from_angles(v.y_offset/core.G3Units.rad, v.x_offset/core.G3Units.rad, 0) for k,v in frame['BolometerProperties'].items()}
+            self.detquat = None
+
+        if frame.type == core.G3FrameType.Scan:
+            timestreams = frame[self.timestreams]
+            bs = numpy.asarray(frame[self.boresight])
+            bs = bs[:,[1,2,3,0]] # Swizzle for TOAST
```
arahlin: ewww.

tskisner:
Sorry this fell through the cracks.  I'll try to look it over in detail 
later today.

mpi/MPIAccumulator.py
```diff
@@ -0,0 +1,101 @@
...
+          ID). This is used only for organization and can be anything that
+          Python can compare. By default, this is SourceName-ObservationID.
+        - Second element is metadata. This information is shared between
+          all processes participating in the communicator at completion and
+          should include any information you need to assess whether the data
+          are interesting for future calculations (for example, the start time
+          of the frame). By default, frames are sorted by this value. None
+          by default.
+        - Third element is the data. This stays node-local. It could be just
+          a copy of the frame (the default) or just the information (in any
```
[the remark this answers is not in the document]
nwhitehorn: It is an actual copy of the frame (though the contents are copy-on-write because of shared pointers).

mpi/MPIAccumulator.py
```diff
@@ -0,0 +1,101 @@
...
+        self.sorter = sorter
+        self.dataframes = dataframes
+        self.localdata = {}
+    def __call__(self, frame):
+        if frame.type == core.G3FrameType.EndProcessing:
+            self.finalize()
+            return
+
+        if frame.type not in self.dataframes:
+            self.extractfunc(frame) # Give it a chance to cache things
```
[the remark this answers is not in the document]
nwhitehorn: Yes!

mpi/MPIAccumulator.py
```diff
@@ -0,0 +1,101 @@
...
+        frame data (metadata, node, data). By default, does no sorting
+        (frames will still appear in a common order across nodes related to
+        the data distribution if unsorted).
+        '''
+        self.mpicomm = mpicomm
+        self.extractfunc = extractfunc
+        def defaultextract(f):
+            if f.type not in self.dataframes:
+                return
+            return ('%s-%d' % (f['SourceName'], f['ObservationID']), None, f)
```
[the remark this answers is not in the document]
nwhitehorn: Yes, changed.

mpi/MPIFileIO.py
nwhitehorn: That is much cleaner, changed.

mpi/MPIFileIO.py
nwhitehorn: Added.

mpi/MPIFileIO.py
nwhitehorn: Done.

mpi/MPIFileIO.py
nwhitehorn: Added

mpi/MPIFileIO.py
nwhitehorn: Done.

mpi/README.rst
nwhitehorn: Fixed.

nwhitehorn:
Ted, are you going to be able to look at this? Or should I merge the non-TOAST parts now?

mpi/TOASTFiller.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def __call__(self, frame):
+        rv = self.mpia(frame)
+        if frame.type != core.G3FrameType.EndProcessing:
+            return rv
+
+        # Build TOAST TOD from our full data set, now that we have it
+        toastobses = []
+        for obs,scans in self.mpia.fullobs.items():
+            nsamptot = sum([i[0]['nsamples'] for i in scans])
```
tskisner: I agree with the general paradigm of the TOASTFiller class, but I think it might be hard to make a general class for this, since every experiment has different toast.TOD derived classes and would need a custom extraction class.

Imagine that an experiment has a bunch of observations, each with lots of frames.  Some external function from the experiment determines how to divide up those frames among some MPI processes.  Now every process runs a G3Pipeline on its frames.  Next:

1.  The experiment defines an "extractor" function that enforces how frames should be unpacked according to naming conventions, etc.

2.  The experiment defines a toast.TOD derived class with some constructor kwargs

3.  A generic "TOASTFiller" class could then take as constructor arguments the extractor class instance, the TOD class type, and the common kwargs.  It would then use MPIAccumulator on the toast group communicator with the specified extractor function and for each observation it would instantiate the experiment TOD class and fill the local cache with the unpacked frame data.

Alternatively, if more specialization was needed, the experiment could just have their own custom TOASTFiller class that did extra things.  I'm not sure if that is needed, but we could try to implement something like this for an experiment and see how it works.

mpi/MPIAccumulator.py
```diff
@@ -0,0 +1,102 @@
...
+        return value for data frames if needed or explicitly return a value
+        for calibration data, if needed. The default extractfunc saves only
+        the frames in the "dataframes" argument.
+
+        sorter is used, if defined, to sort the list of frame data in each
+        group. This is passed to the Python sorted() function as a 'key'
+        argument and receives the three-element tuple stored for each set of
+        frame data (metadata, node, data). By default, does no sorting
+        (frames will still appear in a common order across nodes related to
+        the data distribution if unsorted).
```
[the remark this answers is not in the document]
nwhitehorn: That's basically all correct. Point 3 is generally correct, but off in a minor detail: the process that has the "local" data probably got it from the frame distributor rather than directly from disk itself (there is a swizzle step that allows the set of processes doing IO to be optionally smaller or larger than the number of processes operating on the data).

mpi/TOASTFiller.py
nwhitehorn: This is a tricky one. I think I like your idea, and will take a shot at implementing it. Since the TOASTFiller is experimental anyway, I might merge the rest of this and then poke at it separately, if you don't mind.

PR #13, opened 2019-08-27
Sparse storage for sky maps

nwhitehorn:
Retool the flat-sky and Healpix map storage classes to use (optionally, but by default) a sparse backend for data storage. For non-rectangular flat-sky maps, this can reduce memory consumption by large factors. Substantial parts of the map API are also rewritten to be more useful and consistent. The Healpix map object has also been rewritten from scratch to be less baroque and to support better map storage.

The sparse backends are designed to be transparent to code: the main interface to the maps (operator `[]`/python `__getitem__()`) will allocate pixels in the sparse map on assignment. Operations that require format conversions (`numpy.asarray()`) will implicitly adjust the map storage as needed (to dense storage in this case). C++ and Python functions are also provided to explicitly change the backend. The sparse structures, when used, exist both in RAM and on disk. Maps start sparse by default so that the map-maker can build them gradually without needing the whole sky stored in advance.

For flat-sky projections, two backends are provided: dense (as before) and dense-region sparse. The second stores a contiguous set of columns, each of which has a contiguous set of row entries. This matches the typical structure of a partial-sky map---there is a dense region in the middle surrounded by zeros---without the overhead of storing indices for every pixel. Old flat-sky maps continue to be readable.  FlatSkyMap objects can now be constructed with the projection centered on an arbitrary reference pixel, ensuring that all projections are handled consistently with the WCS prescription.  Each map object is also assigned a `wcs` attribute which can be used for plotting on `WCSAxes`.

Healpix has three backends: dense, indexed-sparse, and ring-sparse. Dense stores every pixel on the sky. Indexed-sparse stores a list of pixel indices and pixels (as a hash map, so log(N) access times). Ring-sparse uses a 2D structure like the flat-sky class in which each ring is a "column" and each pixel on the ring is a "row". The Healpix class has been rewritten from scratch (`HealpixHitPix` etc. are gone) and is not backwards compatible (no cut-sky Healpix maps stored in .g3 files are believed to exist in the wild).

The `G3SkyMap` base class has also been moved from core to coordinateutils, while `quat` and `G3VectorQuat` have been moved into core.


arahlin:
> not explicitly sparse-map related, but maputils.load_skymap_fits barfs when I try to load a fits file made with IDL with an image in the primary HDU (amundsen:/spt/user/tcrawfor/rcw38_150_recalibrated.fits).

I think this should be fixed now.

nwhitehorn:
> Isn't sparse_ NULL here?
Fixed, thanks.

dpdutcher:
3G field mapmaker code works with this now, and I noticed a dramatic decrease in ram usage:
  With the old code making a 0.25 arcminute map, I recorded the mapmaker using a constant 17GB of RAM, then peaking to 33GB during when writing the results to disk. The maps were definitely dominating the ram usage during mapmaking.
  With this update, the constant RAM usage went down to 1.7GB, and peaking at just 2GB when writing to disk.  The ram usage during mapmaking is about what I would expect for being timestream-dominated instead of map-dominated.
  Mapmaking took ~15% longer, which in my test case was an additional 5 minutes.
  The output uncompressed maps are about the same size on disk as the compressed ones.

Basic functionality of the output maps---addition, subtraction, multiply, divide, dimensions, etc.--- are all intact and working, and the "FlatSkyMap - FlatSkyMap = G3SkyMap" bug in the old code has been removed.

tl;dr : Works like it's supposed to and doesn't break anything that I've found!

dpdutcher:
End-to-end test of 3G power spectrum code now agrees identically between this branch and master, too.

mhasself:
Neat work.  I've started looking into how this would fit my use cases.

One thing that I stumbled on immediately is the handling of crpix / crval.  These are important bits of projection information that do not seem to be stored in the FlatSkyMap, and thus are garbled in a read / write cycle.  Compare the two attached maps, in ds9 or similar... identical but the coordinates differ.  The input map was generated with [pixell](https://github.com/simonsobs/pixell).

I am anticipating there will be a similar issue with CEA, though I haven't tested it yet, because I don't see a place for PV2_1 to be carried.

[tan_example.zip](https://github.com/CMB-S4/spt3g_software/files/3622944/tan_example.zip)

tcrawfor:
I had thought that Sasha had replaced our embarrassing home-brew WCS treatment with astropy. If that's not the case, then I'm not surprised something breaks with a projection we rarely use in SPT.

--
Tom Crawford
University of Chicago
Department of Astronomy & Astrophysics
773-702-1564

arahlin:
I did replace the homebrew header handling with the astropy tools, but I the interface to our FlatSkyProjection routines doesn't provide for parameters like PV2_1.  We can certainly add that functionality in.

tcrawfor:
But the lack of PV2_1 functionality doesn't explain why this TAN example breaks, does it? I don't see PV2_1 values in the original header.

--
Tom Crawford
University of Chicago
Department of Astronomy & Astrophysics
773-702-1564

arahlin:
I think it means I probably broke something in the TAN implementation...

mhasself:
ZEA is vulnerable to the same issue as TAN (I just checked).  The implicit assumption is [here]( https://github.com/CMB-S4/spt3g_software/blob/55827da550e3d105102ad44506470c230fc7880f/coordinateutils/python/maputils.py#L440-L444).  The reference point is hard-coded to be the near the center of the map.  But that assumption isn't checked on load.

arahlin:
Ah, the issue here is that 3g software doesn't allow CRPIX to be anywhere but the center of the array.  That's certainly solvable.

tcrawfor:
Taking Daniel's should-be-identical pair of sparse and not-sparse maps, filtering them, finding sources, and comparing those sources to AT20G positions also returns identical results.

One thing that broke that process and had to be fixed manually was the apparent change of the MapPolType attribute from core to coordinateutils. I assume there was a good reason for that, but it will probably break other things too.

mhasself:
Couple more observations:

1. Negative indexing of maps doesn't work, because `size_t` is an unsigned type; e.g. [here](https://github.com/CMB-S4/spt3g_software/blob/55827da550e3d105102ad44506470c230fc7880f/coordinateutils/src/FlatSkyMap.cxx#L604-L609).

2. I think one is not supposed to serialize `size_t` data, because that's a platform-dependent typedef.  I'm talking about: [this](https://github.com/CMB-S4/spt3g_software/blob/55827da550e3d105102ad44506470c230fc7880f/coordinateutils/src/FlatSkyMap.cxx#L108-L109) is [that](https://github.com/CMB-S4/spt3g_software/blob/55827da550e3d105102ad44506470c230fc7880f/coordinateutils/include/coordinateutils/FlatSkyMap.h#L125).

nwhitehorn:
> Couple more observations:
> 
>     1. Negative indexing of maps doesn't work, because `size_t` is an unsigned type; e.g. [here](https://github.com/CMB-S4/spt3g_software/blob/55827da550e3d105102ad44506470c230fc7880f/coordinateutils/src/FlatSkyMap.cxx#L604-L609).
> 
>     2. I think one is not supposed to serialize `size_t` data, because that's a platform-dependent typedef.  I'm talking about: [this](https://github.com/CMB-S4/spt3g_software/blob/55827da550e3d105102ad44506470c230fc7880f/coordinateutils/src/FlatSkyMap.cxx#L108-L109) is [that](https://github.com/CMB-S4/spt3g_software/blob/55827da550e3d105102ad44506470c230fc7880f/coordinateutils/include/coordinateutils/FlatSkyMap.h#L125).

Thanks much, fixed.

arahlin:
Ok, `PV2_1` is now handled consistently for the CEA projection, the reference position can be arbitrary, and a `wcs` attribute is added to `FlatSkyMap` objects for plotting and such.

Also, the test suite verifies numerically consistent transformations with WCS.

PR #14, opened 2019-09-05
G3Reader: support for setting socket timeout.

mhasself:
This isn't as useful as hoped, because the connection needs to be
closed if the timeout expires.  So you can't, for example, set
timeout=0 and include Process() calls in an event loop.  However, the
present solution is still usable in cases where you do not want to
hang on Process() indefinitely and when you are ok with needing to
reconnect the Reader if the timeout expires.

Full non-blocking behavior would be great to have but relies on
getting boost istream_filter to propagate WOULD_BLOCK as distinct from
EOF.  Although I haven't understood the boost stuff fully, one thing
to note is that io::file_descriptor_source definitely does not
check+propagate EAGAIN after calling read(fd).

### DIFF
--- a/core/src/G3Reader.cxx
+++ b/core/src/G3Reader.cxx
@@ -70,12 +70,18 @@ void G3Reader::Process(G3FramePtr frame, std::deque<G3FramePtr> &out)
 		out.push_back(frame);
 	}
 
+	// For python threaded contexts, release the GIL while
+	// blocking for data (which can occur on stream.peek as well
+	// as load(stream_)).  All paths out of this function must
+	// reacquire the lock!
+	PyThreadState *_save = PyEval_SaveThread();
 	if (stream_.peek() == EOF) {
 		if (filename_.size() > 0) {
 			StartFile(filename_.front());
 			filename_.pop_front();
 		} else {
 			// Stop processing
+			PyEval_RestoreThread(_save);
 			return;
 		}
 	}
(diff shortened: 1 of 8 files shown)


core/src/G3Reader.cxx
```diff
@@ -68,12 +70,18 @@ void G3Reader::Process(G3FramePtr frame, std::deque<G3FramePtr> &out)
 		out.push_back(frame);
 	}
 
+	// For python threaded contexts, release the GIL while
+	// blocking for data (which can occur on stream.peek as well
+	// as load(stream_)).  All paths out of this function must
+	// reacquire the lock!
+	PyThreadState *_save = PyEval_SaveThread();
```
nwhitehorn: Is there a way here to tell if we are running with a Python interpreter?

core/src/dataio.cxx
```diff
@@ -18,7 +18,7 @@
 
 void
 g3_istream_from_path(boost::iostreams::filtering_istream &stream,
-    const std::string &path)
+                     const std::string &path, float timeout)
```
nwhitehorn: Could you undo this whitespace change please?

core/src/G3Reader.cxx
mhasself: Yes, there sure is... fixed in https://github.com/CMB-S4/spt3g_software/pull/14/commits/23412a77a26b3ff80708447a8fc209b72057dc33.

core/src/dataio.cxx
mhasself: Fixed in https://github.com/CMB-S4/spt3g_software/pull/14/commits/23412a77a26b3ff80708447a8fc209b72057dc33.

PR #15, opened 2019-10-09
Avoid errors in astronomical calculations due to an out of date IERS database

arahlin:
* By default, attempt to do calculations using the default astropy IERS configuration parameters
 * If the cached file is out of date and the auto download fails, enable extrapolation, with the idea that whatever previously downloaded file is cached is sufficiently new that extrapolation should be fine.
 * If that fails too, use a locally downloaded IERSA file instead, which is then up to the user to maintain up to date.


arahlin:
I think this resolves the two kinds of errors that we see with this module at pole:

(1) if the network is down and the cached file can't be updated but is otherwise intact, just allowing extrapolation will work
(2) if the cached file gets corrupted during download, then falling back to the local file stored with the repository is guaranteed to work, provided that file is sufficiently up to date.

PR #16, opened 2019-10-10
Optimize the Healpix ringsparse representation in memory

arahlin:
The default `ringsparse` representation divides the sphere into rings such that index 0 on a ring corresponds to ra = 0.  This is suboptimal for maps that are centered at ra = 0, like those collected at the South Pole.

This commit adds a boolean optional argument, `shift_ra`, that rearranges the `ringsparse` pixelization such that ring index 0 corresponds to ra = -180 degrees.  This argument defaults to `False`.  It is up to the user to set this option appropriately, either as an argument to the constructor or by setting the `shift_ra` attribute of an existing map object.  When reading in explicitly indexed HEALpix data (either from a fits file or creating a `HealpixSkyMap` instance from `numpy` arrays by hand), this argument is determined empirically from the distribution of input pixels in phi to optimize memory allocation.

For a map with typical South Pole coverage, this improves the memory footprint of the `ringsparse` representation by more than a factor of 2.

The default map representation for new maps or maps loaded from explicitly indexed fits files has also been changed from the `indexedsparse` representation to the `ringsparse` representation.  Finally, setting any of the representation properties to `False` now raises an error, since that is ill-defined behavior.

### DIFF
--- a/coordinateutils/include/coordinateutils/HealpixSkyMap.h
+++ b/coordinateutils/include/coordinateutils/HealpixSkyMap.h
@@ -87,7 +87,7 @@ public:
 	bool IsRingSparse() const { return (ring_sparse_ != NULL); }
 	bool IsIndexedSparse() const { return (indexed_sparse_ != NULL); }
 
-	bool IsShiftRa() const { return shift_ra_; }
+	bool IsRaShifted() const { return shift_ra_; }
 	void SetShiftRa(bool shift);
 
 	class const_iterator {
--- a/coordinateutils/include/coordinateutils/HealpixSkyMap.h
+++ b/coordinateutils/include/coordinateutils/HealpixSkyMap.h
@@ -101,10 +101,12 @@ public:
 		const_iterator(const const_iterator &iter);
 
 		bool operator==(const const_iterator & other) const {
-			return (index_ == other.index_);
+			return map_.ring_sparse_ ? (j_ == other.j_ && k_ == other.k_) :
+			    index_ == other.index_;
 		}
 		bool operator!=(const const_iterator & other) const {
-			return (index_ != other.index_);
+			return map_.ring_sparse_ ? (j_ != other.j_ || k_ != other.k_) :
+			    index_ != other.index_;
 		}
 
 		reference operator*() { return value_; };
--- a/coordinateutils/include/coordinateutils/HealpixSkyMap.h
+++ b/coordinateutils/include/coordinateutils/HealpixSkyMap.h
@@ -30,8 +30,7 @@ public:
  	    bool is_nested = false,
 	    MapCoordReference coord_ref = MapCoordReference::Equatorial,
 	    G3Timestream::TimestreamUnits u = G3Timestream::Tcmb,
-	    G3SkyMap::MapPolType pol_type = MapPolType::None,
-	    bool shift_ra = false);
+	    G3SkyMap::MapPolType pol_type = MapPolType::None);
 
 	HealpixSkyMap();
 	HealpixSkyMap(const HealpixSkyMap& fm);
--- a/coordinateutils/include/coordinateutils/HealpixSkyMap.h
+++ b/coordinateutils/include/coordinateutils/HealpixSkyMap.h
@@ -21,7 +21,8 @@ public:
  	    bool is_nested = false,
 	    MapCoordReference coord_ref = MapCoordReference::Equatorial,
 	    G3Timestream::TimestreamUnits u = G3Timestream::Tcmb,
-	    G3SkyMap::MapPolType pol_type = MapPolType::None);
+	    G3SkyMap::MapPolType pol_type = MapPolType::None,
+	    bool shift_ra = false);
 
 	// Constructor from a numpy array
 	HealpixSkyMap(boost::python::object v,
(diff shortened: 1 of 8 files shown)


coordinateutils/include/coordinateutils/HealpixSkyMap.h
```diff
@@ -86,6 +87,9 @@ class HealpixSkyMap : public G3FrameObject, public G3SkyMap {
 	bool IsRingSparse() const { return (ring_sparse_ != NULL); }
 	bool IsIndexedSparse() const { return (indexed_sparse_ != NULL); }
 
+	bool IsShiftRa() const { return shift_ra_; }
```
nwhitehorn: Maybe "IsRAShifted()"?

coordinateutils/include/coordinateutils/chealpix.h
```diff
@@ -64,13 +64,16 @@ typedef struct {
   long nring;
   double fact1;
   double fact2;
+  int nest;
+  int shift_phi;
```
nwhitehorn: Aren't these actually bools?

coordinateutils/src/HealpixSkyMap.cxx
```diff
@@ -237,6 +256,13 @@ HealpixSkyMap::load(A &ar, unsigned v)
 		ar & make_nvp("data", *indexed_sparse_);
 		break;
 	}
+
+	if (v > 1)
+		ar & make_nvp("shift_ra", shift_ra_);
```
nwhitehorn: Set to false explicitly here if v == 1

coordinateutils/src/HealpixSkyMap.cxx
```diff
@@ -894,6 +920,24 @@ G3SkyMapPtr HealpixSkyMap::Rebin(size_t scale, bool norm) const
 	return out;
 }
 
+void HealpixSkyMap::SetShiftRa(bool shift)
+{
+	assert(!IsRingSparse());
+
```
nwhitehorn: The idea here is that it only works if you aren't already in this representation? The semantics are a little weird that the circumstances under which this can be set (and read) I think are only ones in which this global parameter does nothing. Not quite sure what to do about that, though...

coordinateutils/src/HealpixSkyMap.cxx
```diff
@@ -1034,9 +1072,13 @@ PYBINDINGS("coordinateutils")
 	    .def(bp::init<>())
 	    .add_property("nside", &HealpixSkyMap::nside)
 	    .add_property("nested", &HealpixSkyMap::nested)
+	    .add_property("shift_ra", &HealpixSkyMap::IsShiftRa, HealpixSkyMap_setshiftra,
+		"True if the ringsparse representation of the map is stored "
+		"with the rings centered at ra = 0 deg, rather than ra = 180 deg. "
+		"This property can be changed only when the map is not in the "
+		"ringsparse representation, and may improve ringsparse memory use.")
```
nwhitehorn: Maybe add some more text here: "Although this only affects the map when using ring sparse, this property can only be..."

coordinateutils/src/get_interp_val.c
```diff
@@ -80,9 +80,10 @@ void get_ring_info(map_info *minfo, long iring, long *startpix,
   Allocate and initialize map_info structure.
 
   nside : map dimension
+  shift_phi : if 1, center rings at phi = 0 instead of phi = pi
   populate : if 1, populate all ring parameters
 */
-map_info * init_map_info(size_t nside, int populate) {
+map_info * init_map_info(size_t nside, int nest, int shift_phi, int populate) {
```
nwhitehorn: Same comment about int vs. bool (C99 has bool).

coordinateutils/tests/healpix_maps.py
```diff
@@ -49,13 +49,23 @@
 assert(x2[0] == v0)
 assert(numpy.sum(x2) == numpy.sum(v))
 
+x.shift_ra = True
 x.ringsparse = True # Indexed to ring
 assert(x.nside == 64)
-assert(x.npix_allocated == 1500)
+# assert(x.npix_allocated == 1500)
```
nwhitehorn: Is there a new value to check? I'd like to keep the assert.

coordinateutils/src/HealpixSkyMap.cxx
arahlin: I guess I could make this function explicitly rearrange the ringsparse representation if you change the parameter?  It's a bit of bookkeeping, but doable.  Then it doesn't matter when you set it, and makes switching back and forth to test things out a bit more straightforward.

coordinateutils/include/coordinateutils/chealpix.h
arahlin: That requires enabling C99 in the `cmake` configuration.

coordinateutils/src/HealpixSkyMap.cxx
arahlin: Done

coordinateutils/src/HealpixSkyMap.cxx
arahlin: Removed the text altogether

arahlin:
Oh good catch!  I'll fix that.

coordinateutils/src/get_interp_val.c
arahlin: Since this is the only C code in the library that would require C99, I'm going to just leave things as they are here.  We can revisit if we integrate more C code later on.

arahlin:
In offline discussion, it sounds like the behavior when setting `dense` etc to `False` should be to raise an error rather than do something that's not well defined.  I've implemented that now.

PR #17, opened 2019-10-29
Introducing G3TimesampleMap

mhasself:
This class is essentially G3FrameObjectMap with a .times data
attribute that provides the vector of timestamps for the data in the
Map.  This is useful and efficient in Housekeeping contexts where you
have a bunch of signals that are sampled irregularly but
simultaneously.


core/src/G3Timesample.cxx
```diff
@@ -0,0 +1,182 @@
...
+inline
+int g3_vect_size(const G3FrameObjectPtr vp)
+{
+	auto v = boost::dynamic_pointer_cast<const g3vectype>(vp);
+	return v == nullptr ? -1 : v->size();
+}
+
+template <typename g3vectype>
+inline
+void g3_concat(g3vectype &output, const g3vectype &src1, const g3vectype &src2)
```
arahlin: Maybe we should just add a `Concatenate` method to `G3Vector` types in general?

mhasself:
Thanks... the __setitem__ checking is a good idea.  I didn't think I knew how to do that... but the latest commit adds checking on __setitem__ and set .times.  The result is much more user-proof.  I also cleaned up some const/reference stuff for better performance.

core/src/G3Timesample.cxx
```diff
@@ -5,43 +5,57 @@
...
+		return -1;
+	return vsize;
+}
+
+// Concatenates two compatible vectors and returns the result.
+// Returns nullptr on any incompatibility.
 template <typename g3vectype>
 inline
-G3FrameObjectPtr test_and_concat(const G3FrameObjectPtr src1, const G3FrameObjectPtr src2)
+G3FrameObjectPtr test_and_concat(const G3FrameObjectPtr &src1, const G3FrameObjectPtr &src2)
```
arahlin: Looks like you're mixing pointers and references here and in the previous function.  Should use either `const G3FrameObjectPtr` or `const G3FrameObject &`, but not both.  C++ and I think Boost in particular gets super confused if you do both things at the same time.

core/src/G3Timesample.cxx
```diff
@@ -148,6 +154,48 @@ G3TimesampleMap G3TimesampleMap::Concatenate(const G3TimesampleMap other)
...
+		std::ostringstream s;
+		s << "Cannot set .times because it conflicts with "
+		  << "the established number of samples (" << self.times.size() << ").";
+		throw g3timesample_exception(s.str());
+	}
+	self.times = _times;
+}
+
+static
+G3VectorTime safe_get_times(const G3TimesampleMap &self)
```
arahlin: Should return a reference `const G3VectorTime &` here, maybe?  I am also possibly being too obsessive about memory use.

core/src/G3Timesample.cxx
```diff
@@ -5,43 +5,57 @@
...
 	auto v = boost::dynamic_pointer_cast<const g3vectype>(vp);
 	return v == nullptr ? -1 : v->size();
 }
 
-template <typename g3vectype>
+// Concatenates two compatible vectors.
+template <typename vectype>
 inline
-void g3_concat(g3vectype &output, const g3vectype &src1, const g3vectype &src2)
+vectype vect_concat(const vectype &src1, const vectype &src2)
```
arahlin: Again might want to be careful here with unnecessary copy on return.

core/src/G3Timesample.cxx
mhasself: Passing smart pointers by const reference is well motivated for performance reasons, as it avoids reference counting operations.  This version compiles and passes tests and doesn't leak memory.  It would be surprising if we could break it so easily!

core/src/G3Timesample.cxx
mhasself: Solution found... turns out you can pass a reference to the data member in add_property so that it acts just like (the set part of) def_readwrite!

core/src/G3Timesample.cxx
mhasself: Well, this is an inline function.  But I changed it in latest commit, anyway.

The C++11 solution for this, generally, is "move constructors".  I have no idea whether those are implemented for G3Vector, perhaps through inheritance from std::vector.

core/src/G3Timesample.cxx
arahlin: Interesting. If it works, it works...

arahlin:
Oh one thing I missed: can you `add_spt3g_test` your testing script to the CMakeLists file?

nwhitehorn:
Any outstanding issues to merge this?

arahlin:
Looks good to go to me, but might be good if you had a closer look, @nwhitehorn

PR #22, opened 2019-11-08
Improved project versioning

### LINKED ISSUE #20: Sequential package versions
It would be very useful to have sequential version codes available through the python interface.  In so3g we will often need to require / check for some minimum spt3g version that implements a feature or bug-fix.

I know it's onerous to be constantly tagging and releasing.  If the commit-differential tags like "v0.1.2+23-g7ab341" are available, somehow, then that's sufficient to bridge even large gaps between releases (for my purposes).

arahlin:
* Use `git describe` to store a sequential version string in `spt3g/version.py`
 * Make `spt3g.__version__` a string variable containing the output of `git describe`
 * Also add package version info to `cmake/Spt3gConfigVersion.cmake` for use by derived cmake projects

For this versioning to work properly, the project should be tagged with a git tag of the form "v<major>.<minor>.<patch>".
If the project is not tagged, the `__version__` string falls back to using a git hash, and the cmake `Spt3gConfigVersion.cmake` file is not created.

To check the version in a cmake project, use:

    find_package(Spt3g 0.1.0)

This will check that the Spt3g package has version 0.1.0 or greater.

Versioning also works correctly when building from an exported source tree.   To configure the git repository to automatically populate the version on export, run the following command in the source directory:

    cmake/config_export.sh

And then export the source code with `git archive`.  All of the scripts that populate the version files in the build directory know how to correctly parse the VERSION file in the source tree.

This PR fixes #20


arahlin:
> Looks great. Do you want to also push the "fullversion" string into the PipelineInfo frames?

Good call, done.

arahlin:
Exporting the version with `git archive` should work now.   I added some notes in the main README for setting up the git repository to be able to automatically modify the VERSION file.   Please let me know if this works for anyone other than me...

PR #25, opened 2019-12-31
Combine SPT and PB DAQ scripts into one script

arahlin:
* Optionally import and run (SPT) calibrator, (PB) whwp, syslog logger or GCP watchdog modules
  * Remove extra pb2a daq script that has become out of date


nwhitehorn:
How different is the current one? It would be really good if PB2 did not diverge -- there's a couple of years of improvements to many things that you are missing out on by having a fork.

arahlin:
merged in 0ee43127c

PR #26, opened 2020-01-20
Extract a minimal amount of data from a GCPSlow frame

arahlin:
This allows more efficient processing of arcfile data when no calibration or
re-packaging of data are necessary (e.g. searching for observation or schedule
start/stop times).


nwhitehorn:
Can you add a bit to the docstrings explicitly stating the difference between UnpackTrackerData() and UnpackTrackerMinimal()? For example, "... does the same things as UnpackTrackerMinimal(), with the addition of X,Y,Z".

Otherwise looks great.

arahlin:
Merged in cde777d6

PR #27, opened 2020-01-20
Drop GIL in G3EventBuilder::Process

jlashner:
This PR drops the GIL in the G3EventBuilder::Process function so that rogue can run alongside the G3Pipeline.


nwhitehorn:
This looks great. Can you remove the empty-line deletion at the end?

nwhitehorn:
Also, I think the patch mixes spaces and tabs.

PR #31, opened 2020-03-26
quaternions: ~ operator does conj(), __abs__, and new "/" overloads

mhasself:
For quaternions that express rotations in 3d, the inverse of q =
(a,b,c,d) is q^-1 = (a,-b,-c,-d) and has the effect of "undoing" the
rotation q.  Here we improve access to q^-1 by:

* Defining the ~ operator on quat and G3VectorQuat, as a quick way to
  get conjugates.
* Defining ``__abs__``, so (in Python) ``abs(q)`` returns the Euclidean norm
  sqrt(a^2+b^2+c^2+d^2)
* Overloading the / operator further, to support all combinations of
  double, quat, and G3VectorQuat.

### DIFF
--- a/core/include/core/G3Quat.h
+++ b/core/include/core/G3Quat.h
@@ -33,11 +33,17 @@ double dot3(quat a, quat b);
 
 G3VECTOR_OF(quat, G3VectorQuat);
 
+quat operator ~(quat);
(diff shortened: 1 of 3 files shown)


mhasself:
Wow, thanks for the fast turn-around!  I'll be sending my tax returns to you folks as well.

PR #32, opened 2020-04-13
Refactoring Maps and Mapmaking, Stage 1

arahlin:
This pull request is the first step in migrating generic map handling and binning code from the SPT-3G repo to this public one.  It includes major some major API changes, largely for improved memory handling and also to simplify future work.  After this change, the migration of the map handling code can proceed gradually with minimal changes to the API.

The major changes are:

* Rename `coordinateutils` to `maps` to facilitate incorporating a variety of map-specific functions from the private 3g mapmaker.
* Migrate `RemoveWeights`, `ApplyWeights` and `FlattenPol` functions and pipeline modules to the `maps` module, and greatly improve their memory handling.  These functions now always operate in-place, with no option to do otherwise.  The `RemoveWeights`  module also optionally sets pixels with zero weight to zero instead of `NaN`.  `FlatSkyMap` objects also have a `flat_pol` attribute that is stored with the map object on disk.
* Remove the `WeightType` enum and `weight_type` attribute of the `G3SkyMapWeights` object, and use a `.polarized` attribute instead that checks whether the appropriate matrix elements (QQ, UU, etc) have been assigned.  Also add `.det()` and `.cond()` methods for weights that return maps of the matrix determinant or condition number.
* A new `compact()` method for map and weights objects that removes spurious zero (and optionally `NaN`) pixels.
* `extract_patch()`, `insert_patch()`, `pad()` and `crop()` methods for `FlatSkyMap` objects.
* 2d slicing set/get operations for `FlatSkyMap` objects.
* The `is_weighted`, `is_polarized` and `is_nested` attributes and keywords have been renamed to `weighted`, `polarized`, `nested` for brevity.
* A new `pol_conv` map attribute to keep track of polarization convention.  Issues warnings (for now) if the polarization convention is not set on a U map.
* Modules for extracting map frames from and injecting map frames into pipelines.
* All C++ class method names have been standardized to `CamelCase`.  All python functions and attributes remain `snake_case`.
* Documentation cleanup


maps/include/maps/FlatSkyMap.h
```diff
@@ -55,7 +60,7 @@ class FlatSkyMap : public G3FrameObject, public G3SkyMap {
 	double &operator [] (size_t i) override;
 
 	double at(size_t x, size_t y) const;
-	double operator () (size_t x, size_t y) const { return this->at(x, y); };
+	double operator () (size_t x, size_t y) const {	return this->at(x, y); };
```
nwhitehorn, line 63: What's going on here? Did you replace a space with a space?

maps/include/maps/G3SkyMap.h
```diff
@@ -354,19 +419,19 @@ class G3SkyMapWithWeights : public G3FrameObject {
 	G3SkyMapWeightsPtr RemoveWeights();
 	void ApplyWeights(G3SkyMapWeightsPtr weights);
 
-	StokesVector get_interp_value(double alpha, double delta) const;
-
 	boost::shared_ptr<G3SkyMapWithWeights> Rebin(size_t scale) const;
 
+	void Compress(bool zero_nans = false);
```
nwhitehorn, line 424: I'm not sure "Compress" is the right name for what this does, since it is not what most people think of by "compression". Maybe "Compact"?

maps/python/map_modules.py
```diff
@@ -0,0 +1,162 @@
...
+    'CompressMaps',
+    'RemoveWeights',
+    'ApplyWeights',
+    'FlattenPol',
+    'ConvertTMapsToPolarized',
+    'ConvertPolarizedMapsToT',
+]
+
+@core.indexmod
+def CompressMaps(frame, zero_nans=False):
```
nwhitehorn, line 14: Same nit about "compress" vs. "compact".

maps/src/G3SkyMap.cxx
```diff
@@ -32,13 +36,23 @@ G3SkyMap::serialize(A &ar, unsigned v)
...
+	ar & cereal::make_nvp("weighted", weighted);
+
+	if (v > 2)
+		ar & cereal::make_nvp("pol_conv", pol_conv);
+	else
+		pol_conv = ConvNone;
+
+	if (pol_type == U && pol_conv == ConvNone)
+		log_warn("Map object has pol_type U and unknown pol_conv. "
+			 "Set the pol_conv attribute to IAU or COSMO.");
```
nwhitehorn, line 54: I worry this will be confusing. Serialization and deserialization can happen at non-obvious times, including implicit times (for example, when using an MPI communicator), and it may not be clear from the message what is happening or why. Is this check only for V1 and V2 maps? If so, this should be gated on the version number and the message should probably say something about it being old data. Otherwise, the setter is probably the right place to warn.

maps/src/G3SkyMap.cxx
```diff
@@ -636,6 +850,10 @@ PYBINDINGS("coordinateutils") {
 	      "together.  Returns a new map object.  Map dimensions must be a "
 	      "multiple of the rebinning scale.")
 
+	    .def("compress", &G3SkyMap::Compress, (bp::arg("zero_nans")=false),
```
nwhitehorn, line 853: Same comment again about "compress", which (to me at least) evokes gzip etc.

maps/src/HealpixSkyMap.cxx
```diff
@@ -431,14 +432,57 @@ HealpixSkyMap::ConvertToIndexedSparse()
 	}
 }
 
+void HealpixSkyMap::Compress(bool zero_nans)
```
nwhitehorn, line 435: Compress vs. Compact/etc. again

maps/src/FlatSkyMap.cxx
```diff
@@ -250,14 +260,42 @@ FlatSkyMap::ConvertToSparse()
 	dense_ = NULL;
 }
 
+void FlatSkyMap::Compress(bool zero_nans)
```
nwhitehorn, line 263: Nit about "compress".

maps/python/map_modules.py
```diff
@@ -0,0 +1,162 @@
...
+            m.compress(zero_nans=zero_nans)
+            frame[s] = m
+    return frame
+
+@core.indexmod
+def RemoveWeights(frame, zero_nans=False):
+    '''
+    Remove weights from input maps.  If zero_nans is `True`, empty pixles are
+    skipped and pixels with zero weight are set to 0 instead of NaN.  Operation
+    is performed in place to minimuze memory use.
```
nwhitehorn, line 31: "minimize" instead of "minimuze"

maps/src/G3SkyMap.cxx
```diff
@@ -636,6 +850,10 @@ PYBINDINGS("coordinateutils") {
...
+	      "Convert the map to its default sparse representation, excluding "
+	      "empty pixels, and optionally converting NaN values to zeroes.")
```
nwhitehorn, line 855: Can you include some verbiage here about when you might want/not want to turn NaNs to zeros? E.g. "Converting NaNs to zeroes will reduce memory use, but will remove the distinction in unweighted (or weight-removed) maps between unobserved regions and regions with zero temperature."

maps/include/maps/FlatSkyMap.h
arahlin: no clue.  it may have accidentally been a tab?

maps/include/maps/G3SkyMap.h
arahlin: This and others fixed in 5546cbf

maps/src/G3SkyMap.cxx
arahlin: fixed in 5546cb

maps/src/G3SkyMap.cxx
arahlin: This check is now only done on V1 and V2 maps, and only when the attribute is explicitly set to None for U maps.  The code that requires a polarization convention will rely on the attribute being set appropriately, as in, e.g. the FlattenPol function.  See cb3c3af

PR #33, opened 2020-05-06
In-place Sort() method for G3TimesampleMap objects

arahlin:
This method is useful when the contents of a G3TimestreamMap object are
unordered in time, e.g. when constructed from out-of-order packets.

All sorting operations are done on one vector at a time with replacement, to
minimize the required memory footprint.

### DIFF
--- a/core/src/G3Timesample.cxx
+++ b/core/src/G3Timesample.cxx
@@ -74,17 +74,17 @@ void vect_reorder(vectype &dest, vectype &src, const std::vector<size_t> &idx)
 }
 
 // Reorders elements of a vector, in place.
-// Returns -1 on any incompatibility.
+// Returns false on any incompatibility.
 template <typename g3vectype>
 inline
-int test_and_reorder(G3FrameObjectPtr &src, const std::vector<size_t> &idx)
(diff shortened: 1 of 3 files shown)


core/src/G3Timesample.cxx
```diff
@@ -222,6 +281,8 @@ PYBINDINGS("core")
           "consistency.  Raises ValueError if there are problems.")
 	.def("Concatenate", &G3TimesampleMap::Concatenate,
           "Concatenate two compatible G3TimesampleMap.")
+	.def("Sort", &G3TimesampleMap::Sort,
```
nwhitehorn, line 284: Shouldn't this be sort() instead of Sort() in Python? It differs from the style of the other Python bindings, but sort() is a standard Python function so we probably gain net in terms of POLA.

arahlin: I'm fine with making it lowercase -- I'd also probably make `check` and `concatenate` lowercase in python too, since we already have a lowercase `concatenate` method for G3Timestreams.  @mhasself any opinion here?

mhasself: Lower case interface is fine with Me.  Ahem, with me.

PR #36, opened 2020-07-28
Check units in map arithmetic

arahlin:
Ensure that units match when adding, subtracting or comparing two maps together.
When multiplying or dividing maps, ensure that the unitful map's units get
propagated to the output map.

### DIFF
--- a/maps/src/FlatSkyMap.cxx
+++ b/maps/src/FlatSkyMap.cxx
@@ -397,14 +397,12 @@ G3SkyMap &FlatSkyMap::operator op(const G3SkyMap &rhs) {\
 flatskymap_arithmetic(+=, {}, {g3_assert(units == rhs.units);})
 flatskymap_arithmetic(-=, {}, {g3_assert(units == rhs.units);})
 flatskymap_arithmetic(/=, {ConvertToDense(); (*this->dense_) /= 0.0;},
-    {if (units == G3Timestream::None) units = rhs.units; else if (rhs.units != G3Timestream::None) g3_assert(units == rhs.units);})
+    {if (units == G3Timestream::None) units = rhs.units;})
 
 G3SkyMap &FlatSkyMap::operator *=(const G3SkyMap &rhs) {
 	g3_assert(IsCompatible(rhs));
 	if (units == G3Timestream::None)
 		units = rhs.units;
-	else if (rhs.units != G3Timestream::None)
-		g3_assert(units == rhs.units);
 	try {
 		const FlatSkyMap& b = dynamic_cast<const FlatSkyMap &>(rhs);
 		bool zero = false;
--- a/maps/src/FlatSkyMap.cxx
+++ b/maps/src/FlatSkyMap.cxx
@@ -357,9 +357,10 @@ FlatSkyMap::operator [] (size_t i)
 	return (*this)(i % xpix_, i / xpix_);
 }
 
-#define flatskymap_arithmetic(op, sparsenull) \
+#define flatskymap_arithmetic(op, sparsenull, unitscheck) \
 G3SkyMap &FlatSkyMap::operator op(const G3SkyMap &rhs) {\
 	g3_assert(IsCompatible(rhs)); \
+	unitscheck \
 	try { \
 		const FlatSkyMap& b = dynamic_cast<const FlatSkyMap &>(rhs); \
 		if (dense_) { \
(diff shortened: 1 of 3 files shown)


nwhitehorn:
Does this do the right thing for (map in Tcmb) * (unitless map)? It seems like it produces a map with None units, but it shoould be Tcmb.

marius311:
With this PR:

```
In [6]: uf = FlatSkyMap(rand(10,10), res=1) 
   ...: ul = FlatSkyMap(rand(10,10), res=1, units=None)                                                                             

In [7]: (uf * ul).units                                                                                                             
Out[7]: spt3g.core.G3TimestreamUnits.Tcmb

In [8]: (ul * uf).units                                                                                                             
Out[8]: spt3g.core.G3TimestreamUnits.Tcmb

In [9]: (uf * uf).units                                                                                                             
Out[9]: spt3g.core.G3TimestreamUnits.Tcmb

In [10]: (ul * ul).units                                                                                                            
Out[10]: spt3g.core.G3TimestreamUnits.None
```

>Does this do the right thing for (map in Tcmb) * (unitless map)? It seems like it produces a map with None units, but it shoould be Tcmb.

That's `(uf * ul).units` above and the answer is it produces the right units.

Looking at this again though, `(uf * uf).units` doesn't actually make sense since it should produce something like `Tcmb^2`, but that's not actually a type we have. One could imagine making it an error similar `uf + ul`, but that might break pretty reasonable existing code like

```python
sqrt(sum((m**2) for m in maps))  # compute pixel std-dev 
```

Alternatively, you could also add `Tcmb^2` units. I think it would be fine to do that but stop short of the full rabbit hole of `Tcmb^3`, etc..., since only really variances ever come up. 

Anyway, if this all seems overly pedantic, I'll mention the motivation for it, which is the spt3g <-> CMBLensing.jl autoconversions between maps done by https://github.com/SouthPoleTelescope/CMBLensingSPT3GInterface (this would arise in any interface between these maps, not just limited to CMBLensing.jl). Since the two packages have different unit conventions, one _has_ to know whether a given map is a unitless or unitful quantity to be able to do the conversion. In this regard, its really nice that spt3g is storing the units at all, modulo the minor problems mentioned above.

arahlin:
To be clear, the `units` attribute here is not the units of the map, it's the type of units in which the map is defined.  Weight maps whose `units` attribute is equal to `Tcmb` are in units of inverse temperature squared, weighted Stokes maps are in units of inverse temperature, and unweighted Stokes maps are in units of temperature.  So multiplying two maps together that have the same units type means that the output map also has the same units type (and its up to the user to know order of the conversion factor that is appropriate for each map in question).

This does bring up a good point though, that multiplying two maps maps together that are defined with different types of units is a bit weird, in that it's not clear what the units type attribute should be in that case.  I'm not sure of the right way to handle that, maybe just not allow it?

marius311:
> To be clear, the units attribute here is not the units of the map, it's the type of units in which the map is defined.

marius311:
> To be clear, the units attribute here is not the units of the map, it's the type of units in which the map is defined. Weight maps whose units attribute is equal to Tcmb are in units of inverse temperature squared, weighted Stokes maps are in units of inverse temperature, and unweighted Stokes maps are in units of temperature. 

Yep, fully agree here. 

And to take three examples above, its important to note that weight maps can be identified because they are a `G3SkyMapWeights`, and weighted vs. unweighted maps can be identified as such via `.weighted` attribute. So if I know which of these three it is and if I know the `.units` field (and if I know that in spt3g_sofware temperature is in `G3Units.uK`), I know _exactly_ the physical units I'm dealing with for any of these objects. 

> So multiplying two maps together that have the same units type means that the output map also has the same units type (and its up to the user to know order of the conversion factor that is appropriate for each map in question).

But if `x` is say a `FlatSkyMap(units=Tcmb, weighted=True)`, then `x` and `x * x` are exactly the same type of object and I have nothing like the above to discern them, but they have different physical units. I think this is the fundamental limitation I'm highlighting here.

I think you could solve it by introducing something like a `G3TimestreamUnits.Tcmb^2` which is produced by multiplying two maps together. You could disallow it for `G3SkyMapWeights` where it wouldn't really make sense. All that said, I think could also merge this as-is now and keep a full solution in mind for a future PR.

marius311:
> > So multiplying two maps together that have the same units type means that the output map also has the same units type (and its up to the user to know order of the conversion factor that is appropriate for each map in question).
>
> But if x is say a FlatSkyMap(units=Tcmb, weighted=True), then x and x * x are exactly the same type of object and I have nothing like the above to discern them, but they have different physical units. I think this is the fundamental limitation I'm highlighting here.

Rereading, I guess that's what you mean by "its up to the user to know the order of the factor", ie youre totally aware of this limitation but just think its the user's job. Which is certainly fine, I would just say its nicer if it could be handled automatically (this is made especially obvious by interfacing these two codes which use different units). Anyway, its certainly fine for another time.

nwhitehorn (approved):

PR #37, opened 2020-07-30
make installable on NERSC and document

marius311:
As discussed, this is now installable natively on NERSC and all tests pass.

Would be good if someone verifies they can reproduce this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,7 +51,7 @@ if (SPT3G_VERSION)
 endif()
 
 # Ignore known warnings
-target_compile_options(spt3g INTERFACE -Wno-unknown-warning-option -Wno-unused -Wno-unused-result -Wno-deprecated-register -Wno-sign-compare -Wno-constant-conversion -Wno-self-assign-overloaded -Wno-deprecated-declarations)
+target_compile_options(spt3g INTERFACE -Wno-unknown-warning-option -Wno-unused -Wno-unused-result -Wno-deprecated-register -Wno-sign-compare -Wno-constant-conversion -Wno-self-assign-overloaded -Wno-deprecated-declarations -Wno-maybe-uninitialized)
 
 # Fix bugs in GCC 4.4's strict aliasing code by turning it off
 target_compile_options(spt3g INTERFACE -fno-strict-aliasing)
--- a/cmake/FindFFTW.cmake
+++ b/cmake/FindFFTW.cmake
@@ -8,12 +8,12 @@ if (FFTW_INCLUDES)
 # Already in cache, be silent
 set (FFTW_FIND_QUIETLY TRUE)
 endif (FFTW_INCLUDES)
-find_path (FFTW_INCLUDES fftw3.h)
-find_library (FFTW_LIBRARIES NAMES fftw3)
+find_path (FFTW_INCLUDES fftw3.h HINTS ENV FFTW_INC)
+find_library (FFTW_LIBRARIES NAMES fftw3 HINTS ENV FFTW_DIR)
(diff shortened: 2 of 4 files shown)


arahlin:
Tagging @mhasself for review since I know he's dealt with this issue.

arahlin:
Is it standard for the FFTW and GSL packages to define the environment variables you're using?  If so, we could modify the `cmake/FindFFTW.cmake` and `FindGSL.cmake` modules to include those in their search paths.

marius311:
> Is it standard for the FFTW and GSL packages to define the environment variables you're using? If so, we could modify the cmake/FindFFTW.cmake and FindGSL.cmake modules to include those in their search paths.

I don't think so, I think those are NERSC-module specific choices which wouldn't make sense for be in the cmake files, e.g.:

```
$ module show fftw/3.3.8
-------------------------------------------------------------------
/global/common/software/nersc/cle7/extra_modulefiles/fftw/3.3.8:

conflict	 fftw3 
module-whatis	 FFTW is a C subroutine library for computing the discrete Fourier transform (DFT) in one or more dimensions, of arbitrary input size, and of both real and complex data (as well as of even/odd data, i.e. the discrete cosine/sine transforms or DCT/DST). 
setenv		 FFTW_DIR /usr/common/software/fftw3/3.3.8/gcc/8.2.0/skx/lib 
setenv		 FFTW_INC /usr/common/software/fftw3/3.3.8/gcc/8.2.0/skx/include 
prepend-path	 LD_LIBRARY_PATH /usr/common/software/fftw3/3.3.8/gcc/8.2.0/skx/lib 
prepend-path	 PATH /usr/common/software/fftw3/3.3.8/gcc/8.2.0/skx/bin 
prepend-path	 MANPATH /usr/common/software/fftw3/3.3.8/gcc/8.2.0/skx/share/man 
prepend-path	 CMAKE_PREFIX_PATH /usr/common/software/fftw3/3.3.8/gcc/8.2.0/skx 
-------------------------------------------------------------------
```

marius311:
Well, I guess it couldn't hurt to just add them as one of several environment vars which get searched (assuming you can do that in cmake, which I don't konw)? Then it just works on NERSC without the more complicated cmake command. I think that's a reasonable idea.

arahlin:
Yeah I don’t think it’ll hurt to include those variables in the search path.  I can add that change to this PR in the morning.

CMakeLists.txt
```diff
@@ -51,7 +51,7 @@ if (SPT3G_VERSION)
 endif()
 
 # Ignore known warnings
-target_compile_options(spt3g INTERFACE -Wno-unknown-warning-option -Wno-unused -Wno-unused-result -Wno-deprecated-register -Wno-sign-compare -Wno-constant-conversion -Wno-self-assign-overloaded -Wno-deprecated-declarations)
+target_compile_options(spt3g INTERFACE -Wno-unknown-warning-option -Wno-unused -Wno-unused-result -Wno-deprecated-register -Wno-sign-compare -Wno-constant-conversion -Wno-self-assign-overloaded -Wno-deprecated-declarations -Wno-maybe-uninitialized)
```
nwhitehorn, line 54: What place in the code is this fixing? I'd prefer to fix the code, since this is (often) an actually useful warning being turned off.

marius311: At least one place was I believe this, which is out of our control. https://github.com/boostorg/python/issues/180

marius311:
Just checked with your changes @arahlin and it works without the complicated `cmake` line. I changed the docs to reflect that. Ready to merge.

PR #40, opened 2020-10-06
New map maker

nwhitehorn:
This branch adds a new map binner with a more streamlined interface. The code is broken into four map-binning modules:
* SingleDetectorBoresightBinner: creates (unpolarized) maps for each bolometer in boresight coordinates
* SingleDetectorMapBinner: creates (unpolarized) maps for each bolometer in regular (RA/dec etc.) coordinates
* MapBinner: creates one map from all the detectors (polarized or unpolarized), optionally using OpenMP to parallelize the binning by detector.
* MapMockObserver: creates TOD for all detectors in the BolometerPropertiesMap given an input sky and scan frames containing pointing.

The main goal is to streamline the map-making flow, simplifying configuration and improving performance. Very little that changes any existing code or scripts is actually changed by this patch set -- mostly it just adds modules.

The only real change besides that is the addition of a new core class, G3TimestreamQuat, that is identical to (and inherits from) a G3VectorQuat except that it has some timestream-like properties (start, stop, sample_rate, etc.). This simplifies the use of MapMockObserver and is more generally useful anyway. The coordinate-system modules that make quaternion pointing in maps/ have been updated to emit G3TimestreamQuat instead of G3VectorQuat, and a module (maps.AddTimingToPointingQuats) has been added that updates old data using timing from other objects in the frame. Because G3TimestreamQuat inherits from G3VectorQuat, updates to the data will not break any existing code and all modules except MapMockObserver continue to epect data in the lowest-common-denominator G3VectorQuat and so work with data in either format. Parts of the C++ API that translate between G3Timestream for scaling pointing and quaternions should be updated in future, but have not yet been.


nwhitehorn:
Some example usage doing a round-trip through this code:
```
from spt3g import core, maps
import sys

stub_map = None
for fr in core.G3File(sys.argv[1]):
        if 'T' in fr and fr['Id'] != 'mask':
                print(fr)
                stub_map = fr['T']/fr['Wunpol'].TT
                break

p = core.G3Pipeline()
p.Add(core.G3Reader, filename=sys.argv[2:])
p.Add(lambda f: 'Turnaround' not in f)
p.Add(maps.AddTimingToPointingQuats, key='OnlineRaDecRotation')
if True:
        p.Add(core.Delete, keys=['RawTimestreams_I'])
        p.Add(maps.MapMockObserver, pointing='OnlineRaDecRotation', timestreams='RawTimestreams_I', band=150*core.G3Units.GHz, T=stub_map)
#p.Add(maps.SingleDetectorMapBinner, stub_map=stub_map, pointing='OnlineRaDecRotation', timestreams='RawTimestreams_I')
def synthesize_weights(fr):
        if 'RawTimestreams_I' not in fr:
                return
        w = core.G3MapDouble()
        for i in fr['RawTimestreams_I'].keys():
                w[i] = 1.0
        fr['Weights'] = w
p.Add(synthesize_weights)
p.Add(maps.MapBinner, map_id='Test', stub_map=stub_map, pointing='OnlineRaDecRotation', timestreams='RawTimestreams_I', detector_weights='Weights')
p.Add(core.Dump)
p.Add(lambda f: f.type == core.G3FrameType.PipelineInfo or f.type == core.G3FrameType.Map)
p.Add(core.G3Writer, filename='/tmp/out.g3')

p.Run()

```

maps/src/MapBinner.cxx
```diff
@@ -0,0 +1,290 @@
...
+	for (auto i : *timestreams)
+		dets.push_back(i.first);
+
+	// Clamp num_threads to prevent memory balloon?
+	#pragma omp parallel private(scanT, scanQ, scanU, scanW) shared(chunk_array,nchunks)
+	{
+		#pragma omp single
+		{
+			chunk_array = new struct chunk[omp_get_num_threads()];
+			nchunks = omp_get_num_threads();
```
arahlin, line 178: I think `omp_get_num_threads()` does what you expect outside of the the omp parallel block.  Why can't `chunk_array` and `nchunks` be initialized at lines 163-164?

maps/src/MapBinner.cxx
```diff
@@ -0,0 +1,290 @@
...
+
+		#pragma omp for
+		for (size_t i = 0; i < dets.size(); i++) {
+			BinTimestream(*timestreams->at(dets[i]),
+			    (weights) ? weights->at(dets[i]) : 1,
+			    boloprops_->at(dets[i]), *pointing,
+			    scanT, scanQ, scanU, scanW);
+		}
+	}
+	for (size_t j = 0; j < nchunks; j++) {
```
arahlin, line 202: Might be simpler to just do `T += scanT` in a `pragma omp atomic` block and avoid the chunk array business altogether?

maps/src/MapBinner.cxx
```diff
@@ -0,0 +1,290 @@
...
+	get_detector_pointing(bp.x_offset, bp.y_offset, pointing, T->coord_ref,
+	    alpha, delta);
+
+	auto detpointing = T->AnglesToPixels(alpha, delta);
+
+	if (Q) {
+		// And polarization coupling
+		// XXX: does not handle focal-plane rotation
+		StokesVector pcoupling;
+		set_stokes_coupling(bp.pol_angle, bp.pol_efficiency, pcoupling);
```
arahlin, line 271: Might as well precompute `pcoupling * weight` too.

maps/src/MapMockObserver.cxx
```diff
@@ -0,0 +1,171 @@
...
+public:
+	MapMockObserver(std::string pointing, std::string timestreams,
+	    double band, G3SkyMapConstPtr T, G3SkyMapConstPtr Q,
+	    G3SkyMapConstPtr U, std::string bolo_properties_name);
+	virtual ~MapMockObserver() {}
+
+	void Process(G3FramePtr frame, std::deque<G3FramePtr> &out);
+
+private:
+	void BinTimestream(const G3Timestream &det, double weight,
```
arahlin, line 25: I think this `BinTimestream` method is a copy/paste snafu?

maps/src/SingleDetectorBoresightBinner.cxx
```diff
@@ -0,0 +1,167 @@
...
+		if (template_->units != timestreams->GetUnits())
+			log_fatal("Timestreams have units that do not match "
+			    "earlier timestreams in the pipeline.");
+		// XXX Check for lost detectors here?
+	}
+
+	// Calculate common pointing
+	g3_assert(timestreams->NSamples() == pointing->size());
+#ifdef NOTYET
+	std::vector<size_t> pixels = template_->QuatsToPixels(*pointing);
```
arahlin, line 133: I think this will work if you replace `pointing` -> `pointing * quat(0, 1, 0, 0) / pointing`.

maps/src/pointing.cxx
```diff
@@ -280,7 +280,7 @@ get_fk5_j2000_to_gal_quat()
 
 static void
 create_det_az_el_trans(const G3Timestream &az, const G3Timestream &el,
-    G3VectorQuat &trans_quats)
+    G3VectorQuat &trans_quats) // XXX: switch to G3TimestreamQuat?
```
arahlin, line 283: Why not?

arahlin (changes requested):
Looks good to me, some comments in line.  I'd like to rework the quaternion stuff to make it more efficient, but I'll leave that for a separate PR.

maps/src/FlatSkyMap.cxx
```diff
@@ -354,6 +354,9 @@ FlatSkyMap::at(size_t i) const
 double &
 FlatSkyMap::operator [] (size_t i)
 {
+	if (i == (size_t)-1)
+		return overflow;
+
```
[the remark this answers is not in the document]
nwhitehorn: Are we? We could, of course, but it comes at low cost.

maps/python/quathelpers.py
```diff
@@ -32,10 +32,12 @@ def quat_to_ang(q):
 
 
 @usefulfunc
-def ang_to_quat(alpha, delta):
+def ang_to_quat(alpha, delt, start=None, stop=None):
```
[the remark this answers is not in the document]
nwhitehorn: Whoops... fixed.

maps/src/MapBinner.cxx
nwhitehorn: No, it's OK. It's all part of the same parallel block as below, so omp_get_num_threads() returns (spec'ed and tested) the number of threads participating in the entire parallel block, not just this section.

maps/src/MapBinner.cxx
nwhitehorn: `pragma omp atomic` only really works with scalar/POD types, so this would fail. I could move the same chunk_array stuff into a #pragma omp single block, but that seems basically equivalent to the existing code.

maps/src/MapBinner.cxx
nwhitehorn: Indeed, fixed.

maps/src/pointing.cxx
nwhitehorn: Changing all of these causes API churn in the old mapmaker, which works in terms of G3VectorQuat. I'd kind of like to keep this as-is unless we can rm the old mapmaker, then update it.

maps/src/MapMockObserver.cxx
nwhitehorn: Indeed. Nice catch! Deleted.

maps/src/SingleDetectorBoresightBinner.cxx
nwhitehorn: Do you want this changed? I wasn't sure if you had any additional library methods planned for this since I think you said the API was in flux.

maps/src/pointing.cxx
arahlin: Fair enough.  I'd like to nuke this whole file altogether in the long run...

maps/src/SingleDetectorBoresightBinner.cxx
arahlin: Yeah, I can track the API changes.

maps/src/MapBinner.cxx
arahlin: Sorry, I meant `pragma omp critical` there.

maps/src/FlatSkyMap.cxx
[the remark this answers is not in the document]
arahlin: Secretly using the overflow bin in places gives me the willies.  We left it in for backward compatibility with the old mapmaker and masked filtering code, but I'd rather handle overflow a bit more obviously than this.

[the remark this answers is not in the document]
nwhitehorn: How do you mean "secretly"? The issue here is that AngleToPixel needs to be able to convey that an angle is outside the map, so it has to return *something* (I changed it to -1 in this patch) to indicate that. Once it does that, it only makes sense for operator [] to handle that somewhat gracefully, I thought.

maps/src/MapBinner.cxx
nwhitehorn: Ah, that makes much more sense. That does improve things (change committed).

maps/src/SingleDetectorBoresightBinner.cxx
arahlin: That is, I'd rather remove the NOTYET and clean this block up as part of this PR.  If I end up changing how QuatToPixel works, I can find this block and update accordingly

maps/src/FlatSkyMap.cxx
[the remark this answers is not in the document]
arahlin: You could run through an entire scan, then wonder why your mock timestreams had strange values in them, or why your data didn't show up in your binned map.  This should at least emit a warning or something.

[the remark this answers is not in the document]
nwhitehorn: I'm not sure that's true -- take thumbnail maps, for instance, where this is completely the desired behavior.

maps/src/SingleDetectorBoresightBinner.cxx
nwhitehorn: Done.

maps/src/FlatSkyMap.cxx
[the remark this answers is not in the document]
arahlin: That's fair, I forgot about thumbnails.

maps/src/SingleDetectorMapBinner.cxx
```diff
@@ -0,0 +1,169 @@
...
+			log_fatal("Missing bolometer properties for %s",
+			    det.c_str());
+		get_detector_pointing(bp->second.x_offset, bp->second.y_offset,
+		    *pointing, m->coord_ref, alpha, delta);
+		auto pixels = m->AnglesToPixels(alpha, delta);
+
+		g3_assert(ts->size() == pixels.size());
+		for (size_t j = 0; j < ts->size(); j++) {
+			(*m)[pixels[j]] += (*ts)[j];
+			(*w)[pixels[j]] += 1;
```
arahlin, line 162: Do we want to allow things like PSD weights here?

arahlin (changes requested):

maps/src/SingleDetectorBoresightBinner.cxx
```diff
@@ -0,0 +1,161 @@
...
+#endif
+
+#include <G3Module.h>
+#include <G3Timestream.h>
+#include <G3Quat.h>
+#include <G3Data.h>
+#include <G3Map.h>
+#include <maps/G3SkyMap.h>
+
+// XXX delete this
```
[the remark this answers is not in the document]
nwhitehorn: Indeed. Deleted.

maps/src/SingleDetectorMapBinner.cxx
nwhitehorn: We could -- I'd like to keep feature enhancements for future versions of this code, however.

arahlin (approved):

PR #42, opened 2020-11-16
Change G3VectorInt to wrap vector<int64_t>

mhasself:
This allows G3VectorInt to properly store any valid G3Int, which was
not previously the case.  The G3Vector serialization for int64_t has
been specialized and split (load/save) and is implemented in
G3Vector.cxx.

The unnecessary complication I have added is some basic
compression-on-serialization, where a vector that can be safely
represented as int8/16/32/64 will be stored as that.

Includes a new test and a new "portability" test.


nwhitehorn (approved):
This looks great to me! Thanks a bunch for doing it. I don't have any comments or complaints about the patch. Feel free to merge once Sasha gives her OK as well.

From the SPT side, the only obvious negative impact to me is that this will increase RAM usage for pixel indices when using our old, deprecated map binner. But not by a huge amount and the map binner in question is one we are actively trying to remove, so that doesn't seem like a huge problem.

core/tests/vecint.py
```diff
@@ -0,0 +1,70 @@
...
+        os.remove(test_filename)
+        
+    def test_serialize(self):
+        """Confirm full ranges can be saved and loaded."""
+
+        w = core.G3Writer(test_filename)
+        for isize, val in bit_sizes:
+            f = core.G3Frame()
+            f['v'] = core.G3VectorInt([val] * 10)
+            w.Process(f)
```
arahlin, line 42: It's not common to call the `Process` method directly in python, we usually just treat the pipeline module as a callable function.

arahlin (approved):

PR #44, opened 2020-11-16
OSX Testing Workflow

arahlin:
Fixes for the testing github action with OSX.

Also disable running pipeline modules in a subprocess in python version > 3.7 due to limitations with pickling the module objects, so that tests with the latest python version pass.  Save the correct fix for this for a future PR.

### DIFF
--- a/core/python/modconstruct.py
+++ b/core/python/modconstruct.py
@@ -1,4 +1,4 @@
-from spt3g.core import G3Module, G3Pipeline, G3PipelineInfo, G3Frame, G3FrameType, G3Time, G3ModuleConfig, log_fatal, log_warn
+from spt3g.core import G3Module, G3Pipeline, G3PipelineInfo, G3Frame, G3FrameType, G3Time, G3ModuleConfig, log_fatal, log_error
 try:
     from spt3g.core import multiprocess
     multiproc_avail = True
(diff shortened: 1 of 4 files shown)


core/python/modconstruct.py
```diff
@@ -1,4 +1,4 @@
-from spt3g.core import G3Module, G3Pipeline, G3PipelineInfo, G3Frame, G3FrameType, G3Time, G3ModuleConfig, log_fatal
+from spt3g.core import G3Module, G3Pipeline, G3PipelineInfo, G3Frame, G3FrameType, G3Time, G3ModuleConfig, log_fatal, log_error
 try:
     from spt3g.core import multiprocess
     multiproc_avail = True
```
nwhitehorn, line 4: Why not make multiproc_avail be false for Python > 3.7?

core/tests/multiproc.py
```diff
@@ -6,40 +6,38 @@
...
-	fr['time'] = core.G3Time(int(time.time()*core.G3Units.s))
-	fr['count'] = n
-	n += 1
+    global n
+    if fr.type != core.G3FrameType.Timepoint:
+        return
+    fr['time'] = core.G3Time(int(time.time()*core.G3Units.s))
+    fr['count'] = n
+    n += 1
 m = 0
```
nwhitehorn, line 15: Can you please undo the whitespace changes in this file?

arahlin: as you wish...

core/python/modconstruct.py
arahlin: changed this to raise an `ImportError` in `core/python/multiprocessing.py` instead

PR #48, opened 2021-04-30
G3Writer: release the GIL during IO

mhasself:
This will make life easier in the Simons Observatory Smurf application.

Thanks for the GenerateBlobs api!

### DIFF
--- a/core/src/G3Writer.cxx
+++ b/core/src/G3Writer.cxx
@@ -34,6 +34,16 @@ G3Writer::~G3Writer()
 
 void G3Writer::Process(G3FramePtr frame, std::deque<G3FramePtr> &out)
 {
+        // If in Python context, release the GIL while writing out.
+        // Note we must force serialization of the frames, first,
+        // because in some cases serialization requires calling back
+        // out to Python.
+	PyThreadState *_save = nullptr;
+	if (Py_IsInitialized() && PyGILState_Check()) {
+		frame->GenerateBlobs();
+		_save = PyEval_SaveThread();
+	}
+
 	if (frame->type == G3Frame::EndProcessing)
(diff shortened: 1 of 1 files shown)


core/src/G3Writer.cxx
```diff
@@ -34,6 +34,16 @@ G3Writer::~G3Writer()
 
 void G3Writer::Process(G3FramePtr frame, std::deque<G3FramePtr> &out)
 {
+        // If in Python context, release the GIL while writing out.
+        // Note we must force serialization of the frames, first,
+        // because in some cases serialization requires calling back
+        // out to Python.
+	PyThreadState *_save = nullptr;
```
nwhitehorn, line 41: It looks like you have some mixed tabs and spaces here. Can you please standardize on tabs?

core/src/G3Writer.cxx
```diff
@@ -34,6 +34,16 @@ G3Writer::~G3Writer()
...
+	if (Py_IsInitialized() && PyGILState_Check()) {
+		frame->GenerateBlobs();
```
nwhitehorn, line 43: There's no actual harm to calling GenerateBlobs() unconditionally if you want to simplify this conditional (also no benefit really).

nwhitehorn (changes requested):
Thanks! Some minor comments in the diff. This is known to solve the SMuRF issue?

mhasself:
Fixes as per your comments.  Jack says he can try this on the Smurf server today, and we'll see if it helps.  Indeed it's not clear if our main difficulty is IO or CPU.

nwhitehorn (approved):
Looks to me! Feel free to merge as soon as you know it fixes the problem you have.

PR #49, opened 2021-04-30
Support for boolean/mask maps

nwhitehorn:
It is often useful to have boolean sky maps (e.g. a mask) which are 64x larger than they strictly need to be if represented as doubles. This provides support for such a thing by providing a new sparse storage backend for maps based on std::vector<bool>, which stores entries as single bits. Semantics are the following:
- Comparison operators return a boolean map
- Multiplication of a map by a boolean map is implemented efficiently

To Do:
- [ ] Sparse backend for healpix masks
- [x] Comparison operators as members of G3SkyMap (use `boost::shared_ptr`)
- [x] Efficient mask multiplication operators for FlatSkyMap and HealpixSkyMap
- [x] Tests
- [x] Use masks everywhere
- [x] Create masks from numpy arrays
- [ ] Convert mask to numpy array
- [x] Support numpy ufunc interface for `all()` and friends, at least in the simple case
- [x] Documentation


arahlin:
Ran into a bug accessing the mask `parent` attribute:

```
In [1]: from spt3g import core, maps

In [2]: x = maps.FlatSkyMap(300, 300, core.G3Units.arcmin)

In [3]: m = x.to_mask()

In [4]: m.parent
TypeError                                 Traceback (most recent call last)
Input In [4], in <module>
----> 1 m.parent

TypeError: No to_python (by-value) converter found for C++ type: boost::shared_ptr<G3SkyMap const>
```

arahlin:
This PR is now feature-complete, up to the bug above.

arahlin:
There are many instances of the numpy ufunc reduction wrappers (e.g. `np.all()`) being used for `all`, `any` or `sum` on sky maps or masks.  These now fail because the corresponding C++ methods don't have the signature that the numpy ufunc machinery expects.  I think the right solution is to add wrapper functions to `skymapaddons.py` that allow only the simple case (where the input arguments are `axis=None, out=None`) and otherwise raise a useful error.  I might also consider adding similar functionality for `mean` and `std`, since these are also commonly used.  The alternative is removing the C++ functions and letting the numpy accessors make the maps dense in order to use its internal ufunc machinery, and I'd rather avoid that if possible.  Thoughts?

nwhitehorn:
> I think the right solution is to add wrapper functions to `skymapaddons.py` that allow only the simple case (where the input arguments are `axis=None, out=None`) and otherwise raise a useful error.

This is the way. Anything else will be a lot of work and still be fragile.

nwhitehorn (commented):
Sasha's changes look good to me. I can't actually press "approve" because I apparently wrote too much of the code in the PR but I approved all the same.

PR #51, opened 2021-06-24
Add option to suppress map weight object in map binner output.

nwhitehorn:
This can be useful for e.g. mock observing where the weight matrix
should be the same for every one of many maps and so is known a priori.
For polarized maps, suppressing it can reduce RAM usage by a factor of
four.

### DIFF
--- a/maps/src/MapBinner.cxx
+++ b/maps/src/MapBinner.cxx
@@ -16,7 +16,8 @@ class MapBinner : public G3Module {
 public:
 	MapBinner(std::string output_map_id, const G3SkyMap &stub_map,
 	    std::string pointing, std::string timestreams,
-	    std::string detector_weights, std::string bolo_properties_name);
+	    std::string detector_weights, std::string bolo_properties_name,
+	    bool store_weight_map);
 	virtual ~MapBinner() {}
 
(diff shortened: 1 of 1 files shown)


maps/src/MapBinner.cxx
```diff
@@ -263,12 +271,14 @@ MapBinner::BinTimestream(const G3Timestream &det, double weight,
 			(*T)[detpointing[i]] += pcoupling.t*det[i];
 			(*Q)[detpointing[i]] += pcoupling.q*det[i];
 			(*U)[detpointing[i]] += pcoupling.u*det[i];
-			(*W)[detpointing[i]] += mueller;
+			if (W)
+				(*W)[detpointing[i]] += mueller;
```
arahlin, line 275: Should we even bother populating the `mueller` variable if it's not being stored?

arahlin (approved):
Looks fine to me.  Just one question, mainly for optimization.

maps/src/MapBinner.cxx
nwhitehorn: Good point. The compiler should actually optimize that out, but we can adjust.

PR #52, opened 2021-07-09
MapBinner optimization

arahlin:
* Clear map objects once the output frame is populated to optimize memory usage
 * Avoid computing Mueller matrix elements when map weights are not computed


arahlin:
This code has been tested on the SPT-3G private repo and appears to work as intended.

PR #54, opened 2021-07-21
Consistent handling of `shift_ra` attribute for `HealpixSkyMap` objects

arahlin:
Ensure that all constructors have the same signature.  Also add some tests to check that attributes are properly copied on `clone()`.


arahlin (commented):
This code appears to work and tests pass, so given that Nathan is indisposed I'll just merge it.

PR #55, opened 2021-07-21
Add mass units

arahlin:


nwhitehorn:
Not that I have real objection to the patch, but what is this used for?

arahlin:
> Not that I have real objection to the patch, but what is this used for?

Srini’s foreground sims code defines h and c and k constants for computing blackbody spectra etc.  Those constants have mass units in them.

I wonder if we might also want to just add a bunch of unitful constants like those, so we don’t have to redefine them all over the place.

nwhitehorn (approved):

nwhitehorn:
Makes sense, thanks. I think constants would be great.

PR #56, opened 2021-07-23
Improve mapmaking effiiciency with OpenMP

arahlin:
Ensure that the root thread always writes directly to the main output map.  This
avoids unnecessary map operations and memory allocation in single-threaded
operation (most common with HTC computing), and otherwise provides some speed
and memory improvements for multi-threaded operation.

### DIFF
--- a/maps/src/MapBinner.cxx
+++ b/maps/src/MapBinner.cxx
@@ -179,14 +179,26 @@ MapBinner::Process(G3FramePtr frame, std::deque<G3FramePtr> &out)
 	{
 		G3SkyMapPtr scanT, scanQ, scanU;
 		G3SkyMapWeightsPtr scanW;
- 
-		scanT = T_->Clone(false);
-		if (Q_)
-			scanQ = Q_->Clone(false);
-		if (U_)
-			scanU = U_->Clone(false);
-		if (map_weights_)
-			scanW = map_weights_->Clone(false);
+
+		// root thread writes directly to root map, others write to local clones
+		// should be more efficient (in both memory and time) for any number of threads
+		if (omp_get_thread_num() == 0) {
+			scanT = T_;
+			if (Q_)
+				scanQ = Q_;
+			if (U_)
+				scanU = U_;
+			if (map_weights_)
+				scanW = map_weights_;
+		} else {
+			scanT = T_->Clone(false);
+			if (Q_)
+				scanQ = Q_->Clone(false);
+			if (U_)
+				scanU = U_->Clone(false);
+			if (map_weights_)
+				scanW = map_weights_->Clone(false);
+		}
 
 		#pragma omp for
 		for (size_t i = 0; i < dets.size(); i++) {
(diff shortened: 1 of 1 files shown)


nwhitehorn:
Do you have any measurement of the size of the gains here? They seemed very unlikely to be large when I wrote this code originally and I can imagine cases involving large-area maps where this patch would actually cause a performance regression due to heavier-weight realloc() calls when expanding the map during the scan -- adding rows to a map usually requires copying the entire map in memory and that is a lot more of a penalty with the whole in-progress map than with the little segment, and the cost of that is the same for one row (each sample can have several) or a whole chunk in the gather step.

arahlin:
For the test I ran (one downsampled field scan of the ra0hdec-44.75 SPT-3G field), I found that this fix improved the runtime by about 30% -- 1700 seconds vs 2200 seconds without the fix, both with `OMP_NUM_THREADS=1` -- and did not have a noticeable effect on memory performance.  Running with `OMP_NUM_THREADS` unset (which I believe used 16 threads) caused the memory usage to balloon by 3x, and had a runtime of 2600 seconds, so I imagine the gains are limited for large number of threads.

Is the memory reallocation actually a problem for our sparse map implementation?  If so, wouldn't the problem still exist when doing the critical coadd?

nwhitehorn:
Thanks for the numbers! Is that wall-clock time or CPU time?

The realloc() issue is, in principle, better for the striped version because it has to realloc() the stripe repeatedly, which is smaller, and the big map only once, so you amortize the cost. There are some countervailing pressures, however, since the width of the focal plane means that, often, you are filling in already-allocated pixels. At least here, those countervailing pressures seem to be dominant.

One more question: if the gain is mostly for the OMP_NUM_THREADS=1 case, maybe it would be simpler to make the existing OPENMP #ifdef also a runtime check? Then just run the non-OpenMP branch if we are being asked not to use OpenMP?

arahlin:
> Is that wall-clock time or CPU time?

That's wall time as reported by the `mprof` utility, but I found it to be pretty consistent over several runs.  I should also note that this is for healpix maps, not flat sky maps, although I think I expect the gains to be similar.

> if the gain is mostly for the OMP_NUM_THREADS=1 case, maybe it would be simpler to make the existing OPENMP #ifdef also a runtime check? Then just run the non-OpenMP branch if we are being asked not to use OpenMP?

I expect we'd see some gains for small numbers of threads (say <5, which is not unusual for personal laptops or HTC/HPC hybrid jobs), but I'm not opposed to just making the OPENMP check a runtime thing instead of this fix.

I think we can probably improve things further by, e.g. sorting the detectors into threads by their elevation offsets.  Perhaps an input argument that includes a (manually assembled) sorted list of detectors, that defaults to just using the timestream map keys?

Also, I wonder if an additional problem is just the overhead of computing the pixel mapping.  It's entirely plausible that the sparse healpix map implementation is fairly inefficient in terms of CPU usage.

nwhitehorn (approved):

PR #58, opened 2021-07-27
Add G3Constants namespace and python bindings

arahlin:
This commit provides a set of common mathematical and physical constants, stored
in G3Units, for use in creating code with consistent units handling.

### DIFF
--- a/core/include/core/G3Constants.h
+++ b/core/include/core/G3Constants.h
@@ -5,32 +5,25 @@
 #include <G3Units.h>
 
 namespace G3Constants {
-	using namespace G3Units;
-
-	/* Math */
-	const double pi = M_PI;
-	const double twopi = 2.0 * M_PI;
-	const double halfpi = M_PI / 2.0;
-
 	/* Speed of light */
-	const double c = 299792458.0 * m / s;
+	const double c = 299792458.0 * G3Units::m / G3Units::s;
 
 	/* Planck constant */
-	const double h = 6.62607015e-34 * W * s * s;
+	const double h = 6.62607015e-34 * G3Units::W * G3Units::s * G3Units::s;
 	const double hbar = h / 2.0 / pi;
 
 	/* Boltzmann constant */
-	const double k = 1.380649e-23 * W * s / K;
+	const double k = 1.380649e-23 * G3Units::W * G3Units::s / G3Units::K;
 	const double kb = k;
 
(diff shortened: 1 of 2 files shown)


core/include/core/G3Constants.h
```diff
@@ -0,0 +1,36 @@
...
+#include <G3Units.h>
+
+namespace G3Constants {
+	using namespace G3Units;
+
+	/* Math */
+	const double pi = M_PI;
+	const double twopi = 2.0 * M_PI;
+	const double halfpi = M_PI / 2.0;
+
```
nwhitehorn, line 14: These are already defined constants (even the last, e.g. M_PI_2) in C. Is there a good reason to have private names for them?

core/include/core/G3Constants.h
```diff
@@ -0,0 +1,36 @@
+#ifndef _G3_CONSTANTS_H
+#define _G3_CONSTANTS_H
+
+#include <cmath>
+#include <G3Units.h>
+
+namespace G3Constants {
+	using namespace G3Units;
```
nwhitehorn, line 8: Does this end up making G3Constants::K work by accident?

core/include/core/G3Constants.h
arahlin: No real reason other than completeness, although the list isn't really complete anyway.  I'm fine with removing them.

core/include/core/G3Constants.h
arahlin: Hmm possibly, haven't checked.  Is there a more correct way to do this, other than having to explicitly spell out `G3Units::` everywhere?

core/include/core/G3Constants.h
nwhitehorn: I think I would prefer to drop them if you don't mind.

core/include/core/G3Constants.h
nwhitehorn: After a quick test, it does make G3Constants::K work. I am not aware of an alternative except spelling out G3Units::.

core/include/core/G3Constants.h
arahlin: Done.

core/include/core/G3Constants.h
arahlin: Okie doke

nwhitehorn (approved):

PR #59, opened 2021-07-29
Include G3Units.h only where necessary

arahlin:
This PR removes the `G3Units.h` header dependency from `G3.h`, and instead requires that it be explicitly `#include`'ed in all necessary source files.  This avoids having to rebuild the entire code base every time this header file is changed.

### DIFF
--- a/calibration/src/BoloProperties.cxx
+++ b/calibration/src/BoloProperties.cxx
@@ -1,4 +1,5 @@
 #include <pybindings.h>
+#include <G3Units.h>
 #include <calibration/BoloProperties.h>
 #include <container_pybindings.h>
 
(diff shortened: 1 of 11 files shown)


nwhitehorn (approved):
I don't have any real objections to the patch. I also don't really see the upside -- rebuilding the software takes ~10 minutes and there have been only four changes to G3Units.h since 2016 so the impact seems pretty low.

arahlin:
We've had a few updates to G3Units.h recently, and I think I saw more a few more gaps that we'll likely want to plug eventually.  I found it _just_ annoying enough to deal with rebuilding the whole tree, especially when I'm switching back and forth between branches working on other things, that's I'd rather just fix this...

PR #63, opened 2021-08-21
Improvements to MapMockObserver

arahlin:
This PR ports several simulation features from the SPT-3G private repository into the `MapMockObserver` module.  Namely:

 * An option to sample the input map by interpolation
 * ~An option to construct the output timestreams using a `G3VectorString` of detector names found in the same frame.~ Removed.

### DIFF
--- a/maps/python/quathelpers.py
+++ b/maps/python/quathelpers.py
@@ -78,6 +78,9 @@ def AddTimingToPointingQuats(fr, key, timing_ref='RawBoresightAz'):
     if key not in fr:
         return
 
+    if isinstance(fr[key], G3TimestreamQuat):
+        return
+
     x = fr.pop(key)
     x = G3TimestreamQuat(x)
     x.start = fr[timing_ref].start
--- a/maps/src/MapMockObserver.cxx
+++ b/maps/src/MapMockObserver.cxx
@@ -18,7 +18,7 @@ public:
 	MapMockObserver(std::string pointing, std::string timestreams,
 	    double band, G3SkyMapConstPtr T, G3SkyMapConstPtr Q,
 	    G3SkyMapConstPtr U, std::string bolo_properties_name,
-	    bool interp);
+	    std::string bolos, bool interp);
 	virtual ~MapMockObserver() {}
 
 	void Process(G3FramePtr frame, std::deque<G3FramePtr> &out);
@@ -31,6 +31,7 @@ private:
 
 	std::string boloprops_name_;
 	BolometerPropertiesMapConstPtr boloprops_;
+	std::string bolos_;
 
 	bool interp_;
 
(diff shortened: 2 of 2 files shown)


maps/python/quathelpers.py
```diff
@@ -78,6 +78,9 @@ def AddTimingToPointingQuats(fr, key, timing_ref='RawBoresightAz'):
     if key not in fr:
         return
 
+    if isinstance(fr[key], G3TimestreamQuat):
+        return
+
```
nwhitehorn, line 83: This is a good change, but maybe should go in separately to this PR?

maps/src/MapMockObserver.cxx
```diff
@@ -17,7 +17,8 @@ class MapMockObserver : public G3Module {
 public:
 	MapMockObserver(std::string pointing, std::string timestreams,
 	    double band, G3SkyMapConstPtr T, G3SkyMapConstPtr Q,
-	    G3SkyMapConstPtr U, std::string bolo_properties_name);
+	    G3SkyMapConstPtr U, std::string bolo_properties_name,
+	    std::string bolos, bool interp);
```
nwhitehorn, line 21: Shouldn't "bolos" be a list/vector of strings rather than one string?

nwhitehorn (changes requested):

maps/src/MapMockObserver.cxx
arahlin: It's a key to a `G3VectorString` object that's stored in each Scan frame, presumably because the list can change frame-by-frame?

maps/python/quathelpers.py
arahlin: I added it here since this shim module is only necessary in conjunction with the mock observing module...

maps/src/MapMockObserver.cxx
nwhitehorn: Ah -- that isn't in the docs. Why would it change frame-by-frame? It doesn't in the DAQ, which is what this is trying to simulate.

arahlin: The simstubs have a key in them that's constructed from some timestream map that is put together after a few timestream analysis steps (mainly glitch finding, it seems).  I wouldn't expect the bolometer list to change between frames, but I also don't know why the simstubs would be written this way if that field is completely duplicated...

nwhitehorn: It's an archaism, I think, from the very original version of the map code. The more I think about it, the more I think I would like to leave it out -- unless the number of glitchy detectors is enormous, in which case the observation is presumably cut, simulating them doesn't seem like a huge waste of CPU and it matches the simulation outputs more closely to the real instrument, while making the stubs simpler at the same time.

arahlin: Agreed.  I'll remove this feature and update the mapmaker script appropriately.

nwhitehorn (approved):
Nice! And thanks for reformatting the docstring too -- that's way better.

PR #69, opened 2021-12-17
Add G3Frame JSON serialization

cozzyd:
I envision it being useful to access data in .g3 format from web
applications for monitoring purposes. For that reason, it would
be useful to have a way to convert any data to JSON (one possible alternative
would be a .g3 binary file reader in javascript, but that is much more
work).

Fortunately, cereal supports JSON as an archive format, so adding
JSON output is nearly trivial. We mostly just need to make sure
the JSONOutputArchive version of all the serializations are compiled,
and write a saveJSON() method for g3frame. An asJSON() (as_json() in
Python) method is also provided that returns a string.

There there are a few small differences from binary output:

 - Don't bother emitting crc sums
 - Don't FLAC encode, ever
 - Output the character instead of the number for frametype

Currently, this is only enabled by a new compile-time option (the cmake
variable ENABLE_JSON_OUTPUT), though in the future it will probably be
enabled by default if it doesn't break anything. It does add moderately
to binary size and compile time, but hopefully that's not a huge deal.
The asJSON/as_json methods still exist without JSON support, but return
an error in JSON format.

Also included is a new script, spt3g-jsonify, that will read in a
.g3[.gz] file and output a json stream as a proof of concept.

There are a few places where some other code had to be modified, due to
the different API for binary in cereal text and binary archive formats.
Actually, it's in code that will never be run, but gets generated and
must compile.

Still TODO: 
  - docs
  - tests
  - example HTTP endpoint (likely using boost::beast or equivalent to produce gzipped json).


CMakeLists.txt
```diff
@@ -83,6 +83,15 @@ execute_process(COMMAND mkdir -p ${CMAKE_LIBRARY_OUTPUT_DIRECTORY})
 execute_process(COMMAND mkdir -p ${CMAKE_RUNTIME_OUTPUT_DIRECTORY})
 execute_process(COMMAND ln -fsn ${CMAKE_SOURCE_DIR}/cmake/init.py ${CMAKE_LIBRARY_OUTPUT_DIRECTORY}/__init__.py)
 
+# build options
+set(ENABLE_JSON_OUTPUT "FALSE" CACHE BOOL "Enable JSON output of frames (adds to compile time and binary size)")
```
arahlin, line 87: This build option and compiler flag should just be in core/CMakeLists.txt, otherwise this will trigger rebuilding all of the other packages from scratch.

arahlin (changes requested):

CMakeLists.txt
cozzyd: Hmm, but there are serializable things outside of there too aren't there? (e.g. G3SkyMap looks like it's serializable).

arahlin: Yeah, but the other libraries are built against core, so wouldn’t they just inherit the right properties when linked?  I could be wrong here.

cozzyd: I think the problem is that the packages would then see different versions of serialization.h and the non-core packages wouldn't generate the json serialization code at all (which is done at compile-time). 

Indeed, if I try to do this, everything builds and the tests all pass, but no evidence of json serialization:
```
$ objdump -D libspt3g-core.so | grep rapidjson | wc -l
5168
$ objdump -D libspt3g-maps.so | grep rapidjson | wc -l
0

```
and serializing a frame containing one of those to JSON fails (though less miserably than I expected... I expected to see a symbol lookup error!):

```
In [1]: from spt3g.core import G3Frame      

In [2]: frame = G3Frame()                   

In [3]: from spt3g.maps import FlatSkyMap   

In [4]: frame["map"] = FlatSkyMap()               
...
In [8]: frame.as_json()                     
RuntimeError                              Traceback (most recent call last)
<ipython-input-8-ad9ef20ce3cd> in <module>
----> 1 frame.as_json()

RuntimeError: Trying to save an unregistered polymorphic type (FlatSkyMap).
Make sure your type is registered with CEREAL_REGISTER_TYPE and that the archive you are using was included (and registered with CEREAL_REGISTER_ARCHIVE) prior to calling CEREAL_REGISTER_TYPE.
If your type is already registered and you still see this error, you may need to use CEREAL_REGISTER_DYNAMIC_INIT.
```

arahlin: I see, you’re right.  I still think the definition should be in core/CMakeLists.h since all the code changes are in the core library...  I think there’s a target interface option you can use instead of a global one?  That should properly propagate the definitions to libraries linked against core.

cozzyd: Ah, yes, that does look like it would work! Thanks, I didn't know about that cmake feature.

arahlin:
By the way, I cherry-picked your cmake action fix onto master.  Thanks for fixing that!

arahlin:
I think it probably makes some sense to move the python GIL / threading context machinery to a separate PR, since it's used in a few different places (G3PipelineInfo, G3Reader, G3Writer, G3EventBuilder...) and is not specific to this particular feature.

cozzyd:
Hmm, I wonder why it worked for me without updating the c++ standard...

arahlin:
Ok, you should be able to merge with master, so that just your json changes would be part of this PR now!

arahlin:
@cozzyd I'm tempted to merge the JSON feature into the master branch, and perhaps move the server to a separate repo that uses this package as a dependency.  What do you think?

cozzyd:
Hey! Thanks for doing all the annoying merge work on this!  I think that makes sense from a maintenance perspective. The server was mostly meant to be an example but I neve rfound time to finish it :(.  Feel free to just delete the files here, they won't disappear from git history and can be resurrected later). 

On a separate note, do you know what the future of this git organization will be?

arahlin:
Good question -- this repo is public, but perhaps this is a broader question for "admin".  One option is move this repo to the SouthPoleTelescope organization (under a different name)...

arahlin:
Attaching the working server code here for posterity, just in case.

[serve.tar.gz](https://github.com/user-attachments/files/30313362/serve.tar.gz)

arahlin:
I'm not sure if it makes sense to keep this feature off by default.  Since it's part of the cereal library, I almost want to just get rid of the `WITH_JSON` cmake variable altogether and just leave it on unconditionally.  The change in compile time and binary size isn't that significant.  Thoughts?

arahlin (approved):

nwhitehorn:
> I'm not sure if it makes sense to keep this feature off by default. Since it's part of the cereal library, I almost want to just get rid of the `WITH_JSON` cmake variable altogether and just leave it on unconditionally. The change in compile time and binary size isn't that significant. Thoughts?

I would make it unconditionally on.

core/include/core/serialization.h
```diff
@@ -14,14 +17,39 @@
 #include <pybindings.h>
 #include <G3Logging.h>
 
-#define G3_SERIALIZABLE_CODE(x) \
+#define G3_SERIALIZABLE_CODE_BINARY(x) \
 template void x::serialize(cereal::PortableBinaryOutputArchive &, unsigned); \
 template void x::serialize(cereal::PortableBinaryInputArchive &, unsigned); \
 
-#define G3_SPLIT_SERIALIZABLE_CODE(x) \
+#define G3_SPLIT_SERIALIZABLE_CODE_BINARY(x) \
```
nwhitehorn: For compat, I think it is likely better not to rename this.

core/include/core/serialization.h
```diff
@@ -14,14 +17,39 @@
...
 template void x::save(cereal::PortableBinaryOutputArchive &, unsigned) const; \
 template void x::load(cereal::PortableBinaryInputArchive &, unsigned); \
 
+#ifdef G3_WITH_JSON
+
+#define G3_SERIALIZABLE_CODE_JSON(x) \
+template void x::serialize(cereal::JSONOutputArchive &, unsigned); \
+template void x::serialize(cereal::JSONInputArchive &, unsigned); \
```
nwhitehorn, line 22: Why not just add these specializations to the non-JSON #defines? That avoids modifying every class.

core/src/G3Timestream.cxx
```diff
@@ -437,6 +437,111 @@ template <class A> void G3Timestream::load(A &ar, unsigned v)
 	}
 }
 
+#ifdef G3_WITH_JSON
+template <> void G3Timestream::save(cereal::JSONOutputArchive &ar, unsigned v) const
```
nwhitehorn: Why doesn't the regular save method work? Is the goal just not to do FLAC?

maps/src/G3SkyMap.cxx
```diff
@@ -25,7 +25,7 @@ G3SkyMap::serialize(A &ar, unsigned v)
 		std::vector<double> dat;
 		ar & cereal::make_nvp("data", dat);
 		
-		uint32_t xpix, ypix;
+		uint32_t xpix = 0, ypix = 0;
```
nwhitehorn, line 28: This feels unrelated?

maps/src/HealpixSkyMap.cxx
```diff
@@ -97,6 +97,8 @@ HealpixSkyMap::load(A &ar, unsigned v)
 		ar & make_nvp("nested", nested);
 	} else {
 		ar & make_nvp("info", info_);
+		nside = info_.nside();
```
nwhitehorn, line 100: This feels unrelated, too?

maps/src/G3SkyMap.cxx
arahlin: This and the fix below showed up as "maybe be used uninitialized" errors when I enabled JSON serialization.  It's possible these are issues "only" with g++-13, but for whatever reason don't seem to be a problem with the binary specializations...

core/src/G3Timestream.cxx
arahlin: I think so.

core/include/core/serialization.h
arahlin: Doesn't that then require modifying every class to include the JSON specializations?  Below I re-define these as the combination of the binary and JSON ones.

core/include/core/serialization.h
arahlin: I redefine G3_SERIALIZABLE_CODE as the combination of G3_SERIALIZABLE_CODE_JSON and G3_SERIALIZABLE_CODE_BINARY below this, for exactly this reason.

core/src/G3Timestream.cxx
arahlin: Actually, it's to be able to read flac-encoded timestreams from binary and write to JSON without the encoding.  The JSON archive format doesn't support a bunch of the specializations that the FLAC encoding code uses, e.g. loadBinary().

arahlin: I tried doing this with type traits and constexpr in a few places, but ended up with unexpected test failures with FLAC encoded binary data...

arahlin: Got that working, actually.

core/include/core/serialization.h
nwhitehorn: It shouldn't. Very few of them know the archive type is PortableBinaryArchive; almost every class (or even every class?) uses templates, and the templated functions should work equally well for both.

core/include/core/serialization.h
nwhitehorn: Why even separate them at all, though? Just have one block with four specializations.

arahlin: For the use case in G3Frame where I have specific specializations for the JSON archives and the usual ones for binary archives.  If you prefer I can just spell out the four template functions by hand there, but until I refactored the timestream functions, there were at least two instances of this kind of construct.

core/include/core/serialization.h
arahlin: Yes, that's why all of the classes that aren't G3Frame call G3_SERIALIZABLE_CODE...

core/include/core/serialization.h
nwhitehorn: But does that cause any problems in those cases? I would not expect it to cause any issues when there are explicit specializations.

core/include/core/serialization.h
nwhitehorn: See other comment about this. Does it actually cause issues in practice?

core/include/core/serialization.h
arahlin: Yeah, I get warnings about specializations defined before initializations or something like that.

arahlin: If I put the instantiation first I get this error:

```
/Users/arahlin/code/spt/spt3g_software_public/core/src/G3Frame.cxx:282:15: error: explicit specialization of 'save<cereal::JSONOutputArchive>' after instantiation
  282 | void G3Frame::save(cereal::JSONOutputArchive &ar, unsigned v) const
      |               ^
/Users/arahlin/code/spt/spt3g_software_public/core/src/G3Frame.cxx:214:1: note: explicit instantiation first required here
  214 | G3_SPLIT_SERIALIZABLE_CODE(G3Frame);
      | ^
```

If I put the specialization first, I get this warning:

```
/Users/arahlin/code/spt/spt3g_software_public/core/src/G3Frame.cxx:475:1: warning: explicit instantiation of 'save<cereal::JSONOutputArchive>' that occurs after an explicit specialization has no effect [-Winstantiation-after-specialization]
  475 | G3_SPLIT_SERIALIZABLE_CODE(G3Frame);
      | ^
```

arahlin: If you like, I can make the G3Frame specializations all explicit, not call any of the macros there, and combine all the instantiations into one macro for the classes.

nwhitehorn: I like that idea. I can do it later today too if you want.

arahlin: Done!

nwhitehorn (approved):

PR #74, opened 2022-03-03
Add gcp.ARCFile class analogous to core.G3File

arahlin:
Optional `extract` keyword argument to also run `gcp.ARCExtract` on each frame.

Also, add a `core.InjectFrame` utility function for injecting arbitrary frames into a pipeline, used with `gcp.ARCFile(..., extract=True)`.


nwhitehorn (approved):
Cute. I like it.

PR #75, opened 2022-03-09
Add support in the MapBinner for per-scan (or sub-obs) map-making.

nwhitehorn:
This adds a new parameter (map_per_scan) to the map binner that can
control when maps are emitted. If defaults to False, preserving existing
behavior. If set to True, or to a callable that returns True, it will
result in the in-progress map being flushed to the data stream. This
enables simple creation of per-scan maps (set to True) or more complex
bundlings (with a callable).

### DIFF
--- a/maps/src/MapBinner.cxx
+++ b/maps/src/MapBinner.cxx
@@ -18,7 +18,7 @@ public:
 	MapBinner(std::string output_map_id, const G3SkyMap &stub_map,
 	    std::string pointing, std::string timestreams,
 	    std::string detector_weights, std::string bolo_properties_name,
-	    bool store_weight_map);
+	    bool store_weight_map, boost::python::object map_per_scan);
 	virtual ~MapBinner() {}
 
 	void Process(G3FramePtr frame, std::deque<G3FramePtr> &out);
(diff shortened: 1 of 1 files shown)


maps/src/MapBinner.cxx
```diff
@@ -90,6 +92,14 @@ EXPORT_G3MODULE("maps", MapBinner,
...
 "    set to False if and only if you know what the weights are a priori \n"
 "    (e.g. mock observing, where they are the same as the data maps).\n"
+"map_per_scan: boolean or callable, optional\n"
+"    Defaults to False. If set to True, will make an output map for every \n"
+"    input Scan frame, which can be useful for fine-grained time-domain \n"
+"    maps. Can also be set to a Python callable for complex situations. In \n"
+"    this case, the callable will be called on each Scan frame and passed \n"
+"    the frame as an argument. If it returns True, the map binner will emit \n"
+"    a map frame with the data since the last emitted map frame but *before* \n"
+"    the current Scan; if False, the binner will continue accumulating data.\n"
```
kferguson42, line 102: In the callable case, doesn't it make more sense to include the current Scan frame instead of emitting only everything that came before?

nwhitehorn: The idea is that this indicates the start of *new* data, which is easier to detect than the end of old data since you don't have to know the future.

PR #76, opened 2022-03-23
Add ReprojectMaps pipeline module to simplify reprojection of entire map frames

arahlin:


arahlin:
Joshua has tested this code and it seems to work fine, for T-only maps at least.

nwhitehorn (approved):

nwhitehorn:
Cute! Might be worth saying explicitly that it also does Healpix<->FlatSky.

PR #80, opened 2022-07-28
Super Timestream: Integrate non-double types and array contiguity to timestreams and timestream maps

nwhitehorn:
This provides most (all of the important ones?) of the features of the SO G3SuperTimestream class in the base G3Timestream/G3TimestreamMap. Highlights of the changes:
 * A G3TimestreamMap can now be treated as a 2D array as a read/write object in numpy.
 * A G3TimestreamMap can be compactified into a contiguous 2D block (necessary for above) when needed.
 * A G3TimestreamMap can be initialized from a 2D numpy array and a list of keys, optionally sharing the numpy array's underlying data buffer.
 * Both raw timestreams and timestream maps now support a variety of integer and floating-point data types.
 * FLAC-compressed timestreams are now deserialized as floats instead of doubles, which can be done without data loss since the bit depth of the significand of an IEEE 754 32-bit floating-point number is identical to the bit depth of FLAC (24 bits).

No breaking changes are made to the API for either G3Timestream or G3TimestreamMap.


arahlin (changes requested):
As discussed on slack, it would be useful for the from-numpy constructor to take an optional argument, `copy_data` to determine whether to copy the numpy array (True, default behavior, consistent with other 3g constructors) or to make it shared with the input numpy array (False).

arahlin (approved):

PR #81, opened 2022-08-02
Add seek and tell functions from so3g.G3IndexedReader

arahlin:
This PR provides an implementation for seeking through a G3 file for positions of particular frames, allowing faster parsing of data after the initial pass.  Most efficient for streams containing many interleaved frame types.

Includes a test of the functionality using two passes through an existing data file, first to cache the desired frame position, and then to seek to that position.

### DIFF
--- a/core/src/G3Reader.cxx
+++ b/core/src/G3Reader.cxx
@@ -10,7 +10,7 @@ G3Reader::G3Reader(std::string filename, int n_frames_to_read,
     n_frames_read_(0), timeout_(timeout)
 {
 	boost::filesystem::path fpath(filename);
-	if (filename.find("://") == -1 &&
+	if (filename.find("://") == std::string::npos &&
 	   (!boost::filesystem::exists(fpath) ||
 	    !boost::filesystem::is_regular_file(fpath)))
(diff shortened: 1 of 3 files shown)


nwhitehorn (approved):

core/src/G3Reader.cxx
```diff
@@ -108,6 +108,14 @@ void G3Reader::Process(G3FramePtr frame, std::deque<G3FramePtr> &out)
 	n_frames_read_++;
 }
 
+int G3Reader::Seek(int offset) {
+	return boost::iostreams::seek(stream_, offset, std::ios_base::beg);
+}
+
+int G3Reader::Tell() {
```
nwhitehorn, line 115: Same here.

core/src/G3Reader.cxx
```diff
@@ -108,6 +108,14 @@ void G3Reader::Process(G3FramePtr frame, std::deque<G3FramePtr> &out)
 	n_frames_read_++;
 }
 
+int G3Reader::Seek(int offset) {
```
nwhitehorn, line 111: This needs to be off_t rather than int.

PR #89, opened 2022-10-04
Area-preserving bilinear interpolation for Healpix maps

arahlin:
The weights computed for bilinear interpolation are computed in (theta, phi) coordinates, such that the effective sample location is shifted north or south compared to its actual location, with a shift amplitude that scales like 1/nside and is largest near the poles.  This PR changes the GetInterpPixelsWeights method to compute weights in (z, phi) coordinates, so that the weights are area-preserving and calculated in the coordinates in which the Healpix pixel boundaries are defined.

Note that this means that `HealpixSkyMap.get_interp_values()` no longer produces the same results as `healpy.get_interp_val()`.

Thank you to Eric Hivon for identifying the issue.

### DIFF
--- a/maps/README.rst
+++ b/maps/README.rst
@@ -88,11 +88,13 @@ Map values can be tested using ``.isnan(), .isinf(), .isfinite()`` methods as we
 Map Interpolation
 =================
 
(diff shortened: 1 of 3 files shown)


hivon (changes requested):
I see one problem with the new code, not affecting SPT (since it is for the southernmost ring): line 364 should be
 double wz = (z - z1) / ( -1.0 - z1);

hivon (changes requested):
Actually, there is similar problem at the north pole of Healpix: line 353 should read
`double wz = (z - 1.0)/(z2 - 1.0)`

yomori (approved):
Tried mock-observing and verified that the code works.

Issue #101, opened 2023-04-13
Cannot use empty G3TimestreamMap() constructor


arahlin:
Building with #95 breaks on the private SouthPoleTelescope/spt3g_software repo:

```
[ 79%] Building CXX object todfilter/CMakeFiles/todfilter.dir/src/fftutils.cpp.o
/github/actions-runner/_work/spt3g_software/spt3g_software/todfilter/src/fftutils.cpp: In function ‘void fft_filter_mem_friendly(const G3TimestreamMap&, const G3VectorComplexDouble&, G3TimestreamMap&, bool, G3MapDoubleConstPtr)’:
/github/actions-runner/_work/spt3g_software/spt3g_software/todfilter/src/fftutils.cpp:307:30: error: use of deleted function ‘G3TimestreamMap& G3TimestreamMap::operator=(const G3TimestreamMap&)’
   out_data = G3TimestreamMap();
                              ^
In file included from /github/actions-runner/_work/spt3g_software/spt3g_software/todfilter/include/todfilter/numericutils.h:9,
                 from /github/actions-runner/_work/spt3g_software/spt3g_software/todfilter/src/fftutils.cpp:3:
/github/actions-runner/_work/spt3g_software/spt3g_software/core/include/core/G3Timestream.h:176:7: note: ‘G3TimestreamMap& G3TimestreamMap::operator=(const G3TimestreamMap&)’ is implicitly declared as deleted because ‘G3TimestreamMap’ declares a move constructor or move assignment operator
 class G3TimestreamMap : public G3FrameObject,
       ^~~~~~~~~~~~~~~
/github/actions-runner/_work/spt3g_software/spt3g_software/todfilter/src/fftutils.cpp: In function ‘void fft_filter_mem_friendly_w_multi_filters(G3TimestreamMap&, G3VectorDouble&, G3MapVectorComplexDouble&, G3TimestreamMap&)’:
/github/actions-runner/_work/spt3g_software/spt3g_software/todfilter/src/fftutils.cpp:433:43: error: use of deleted function ‘G3TimestreamMap& G3TimestreamMap::operator=(const G3TimestreamMap&)’
         filtered_ts_map = G3TimestreamMap();
                                           ^
At global scope:
cc1plus: error: unrecognized command line option ‘-Wno-self-assign-overloaded’ [-Werror]
cc1plus: error: unrecognized command line option ‘-Wno-constant-conversion’ [-Werror]
[...]

PR #102, opened 2023-04-13
Explicitly force G3TimestreamMap to have assignment operators

cnweaver:
This solves the issue in https://github.com/CMB-S4/spt3g_software/issues/101.


arahlin (approved):
Thanks for fixing this!

PR #103, opened 2023-04-17
Timestream serialization speed

cnweaver:
These changes roughly double the speed at which large amounts of integer timestream data can be serialized, and should provide much of that benefit to more common floating-point data, as well.


arahlin (approved):
Awesome!  Does this by chance resolve @tskisner's issue with slow serialization?

We should put some thought into proper compression for the non-int32 types, and allow units other than `Counts`, but that's for a separate PR.

cnweaver:
I think @tskisner will have to speak to how useful this is for his purposes; as noted it doesn't directly do anything to help non-`Counts` data (and also only has an impact when compression is turned on), so it will depend on what he needs to store.

PR #106, opened 2023-05-16
Updates to dfmux library for compatibility with hidfmux IceBoard firmware

arahlin:
* Configurable internal clock rate to ensure that the IceBoard sub-second counter can be converted to a proper timestamp
* IceBoard packets indexed by module and sub-module block
* Add `center_frequency` and `bandwidth` attributes to `BolometerProperties` objects, to differentiate between the actual mm-wave observing frequency of the resonator and the band into which channels would be binned during mapmaking.

### DIFF
--- a/calibration/include/calibration/BoloProperties.h
+++ b/calibration/include/calibration/BoloProperties.h
@@ -13,7 +13,8 @@ class BolometerProperties : public G3FrameObject
 {
 public:
 	BolometerProperties() : x_offset(NAN), y_offset(NAN), band(NAN),
-	    pol_angle(NAN), pol_efficiency(NAN), coupling(Unknown) {}
+	    center_frequency(NAN), bandwidth(NAN), pol_angle(NAN),
+	    pol_efficiency(NAN), coupling(Unknown) {}
 	std::string Description() const;
 
 	std::string physical_name; /* e.g. D4.A2.3.Y */
@@ -24,6 +25,7 @@ public:
 
 	double band;		/* Standard frequency units */
 	double center_frequency;
+	double bandwidth;
 	double pol_angle;	/* Standard angular units */
 	double pol_efficiency;	/* 0-1 */
 
--- a/calibration/include/calibration/BoloProperties.h
+++ b/calibration/include/calibration/BoloProperties.h
@@ -23,6 +23,7 @@ public:
 	double y_offset; /* and el */
 
 	double band;		/* Standard frequency units */
+	double center_frequency;
 	double pol_angle;	/* Standard angular units */
 	double pol_efficiency;	/* 0-1 */
 
@@ -44,7 +45,7 @@ public:
 };
 
 G3_POINTERS(BolometerProperties);
-G3_SERIALIZABLE(BolometerProperties, 6);
+G3_SERIALIZABLE(BolometerProperties, 7);
 
 G3MAP_OF(std::string, BolometerProperties, BolometerPropertiesMap);
 
--- a/dfmux/include/dfmux/DfMuxCollector.h
+++ b/dfmux/include/dfmux/DfMuxCollector.h
@@ -19,6 +19,7 @@ public:
 	int32_t block;          /* Sub-module block number (0-7) */
 	int32_t nmodules;	/* Total number of modules on board (8) */
 	int32_t nblocks;        /* Total number of blocks per module (8) */
+	int32_t nchannels;      /* Total number of channels per block (128) */
 	DfMuxSamplePtr sample;	/* Pointer to the DfMuxSample */
 };
 
--- a/dfmux/include/dfmux/DfMuxCollector.h
+++ b/dfmux/include/dfmux/DfMuxCollector.h
@@ -50,6 +50,10 @@ public:
 	int Start();	// Start listening thread
 	int Stop();	// Stop listening thread
 
+	// Configure iceboard clock rate
+	void SetClockRate(uint64_t rate) { clock_rate_ = rate; };
+	uint64_t GetClockRate() { return clock_rate_; };
+
 private:
 	int SetupUDPSocket(const char *listenaddr);
 	int SetupSCTPSocket(std::vector<std::string> hosts);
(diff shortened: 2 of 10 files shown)


calibration/include/calibration/BoloProperties.h
```diff
@@ -23,6 +23,7 @@ class BolometerProperties : public G3FrameObject
 	double y_offset; /* and el */
 
 	double band;		/* Standard frequency units */
+	double center_frequency;
```
nwhitehorn, line 26: If the SLIM detectors have more properties here (the 3G ones do have this property, but from an implementation point of view), it might make sense to subclass this rather than modifying the class. That also provides an opportunity to rename the subclass if people object to "bolometer" without touching any of the 3G codebase.

dfmux/include/dfmux/DfMuxCollector.h
```diff
@@ -16,7 +16,10 @@ class DfMuxSamplePacket : public G3FrameObject {
 public:
 	int32_t board;		/* Board serial number */
 	int32_t module;		/* Module number (0-7) */
+	int32_t block;          /* Sub-module block number (0-7) */
```
nwhitehorn, line 19: Add a comment that this is always 0 for old-school DfMux?

dfmux/include/dfmux/DfMuxCollector.h
```diff
@@ -16,7 +16,10 @@ class DfMuxSamplePacket : public G3FrameObject {
...
 	int32_t nmodules;	/* Total number of modules on board (8) */
+	int32_t nblocks;        /* Total number of blocks per module (8) */
```
nwhitehorn, line 21: And that this isn't always 8.

dfmux/include/dfmux/DfMuxCollector.h
```diff
@@ -50,6 +53,10 @@ class DfMuxCollector {
 	int Start();	// Start listening thread
 	int Stop();	// Stop listening thread
 
+	// Configure iceboard clock rate
+	void SetClockRate(double rate);
+	double GetClockRate() { return clock_rate_; };
```
nwhitehorn, line 58: Does this get initialized by default to 100 MHz anywhere for old DfMux?

dfmux/src/DfMuxCollector.cxx
```diff
@@ -96,6 +97,10 @@ RawTimestampToTimeCode(RawTimestamp stamp)
 				stamp.y = last_stamp.y;
 		}
 	}
+
+	// fix the subsecond counter if the iceboard is using
+	// an internal clock with a rate other than 100 MHz
+	stamp.ss = htole32(le32toh(stamp.ss) * clock_scale);
```
nwhitehorn, line 103: Does this work if the clock is *faster* than 100 MHz? Won't it introduce round-off errors?

nwhitehorn (changes requested):

calibration/include/calibration/BoloProperties.h
arahlin: It sounds like this is the only additional property, but maybe @adamanderson can confirm.

dfmux/include/dfmux/DfMuxCollector.h
arahlin: Yep, in the constructors.

calibration/include/calibration/BoloProperties.h
nwhitehorn: I was thinking of this basically as an excuse for future-proofing, if you wanted one.

dfmux/src/DfMuxCollector.cxx
arahlin: Empirically it seems to work for 125 MHz clocks.  We should probably do a more thorough check, and/or avoid the round-trip endianness conversion.

dfmux/include/dfmux/DfMuxCollector.h
nwhitehorn: Ah, indeed. Could you add a comment about the default somewhere?

dfmux/src/DfMuxCollector.cxx
nwhitehorn: It's not the endianness that worries me, it's the integer data type. Time deltas of 8 ns (1/125 MHz) are not representable with an integer data type and 3G units. I can't imagine +/- 10 ns jitter is meaningful for your application -- and it wouldn't be at all surprising if the "synchronized" sampling is only good to a few ns to begin with -- but I wanted to flag it here.

calibration/include/calibration/BoloProperties.h
arahlin: I think making a subclass just to add a single property adds too much duplicate boilerplate code.  It's probably easier to (eventually) just rename Bolometer->Detector in a backward-compatible way and be done with it.

dfmux/src/DfMuxCollector.cxx
arahlin: I'll use doubles internally and convert to integer only on return

dfmux/include/dfmux/DfMuxCollector.h
arahlin: Added a comment in the source code and a note in the python bindings for `clock_rate`.

calibration/include/calibration/BoloProperties.h
nwhitehorn: I will point out that the SLIM detectors *are* bolometers, just not TESes.

nwhitehorn (approved):

calibration/include/calibration/BoloProperties.h
adamanderson: Subclassing or hardcoding here is fine with me. There will be a few additional properties added, following discussions with Jessica, Kirit, and Pete.

adamanderson (approved):

calibration/include/calibration/BoloProperties.h
nwhitehorn: What kinds of properties? Center frequency applies to TESes as well, though we conflate it with nominal frequency on SPT. If they are really different/extra things, I think a subclass would make a lot of sense.

PR #112, opened 2023-08-15
Add frame-level bzip2 compression and decompression

arahlin:
This PR integrates bzip2 support to be on par with gzip support for G3 file I/O. Files with the ".g3.bz2" extension will run through the appropriate bz2 algorithm on read and write.  Bzip2 support is optional and automatically detected by cmake, but can be manually overridden using a cmake argument.

This completes the bzip2 handling that was started in PR #97.


cnweaver (approved):
This looks reasonable to me, with the caveat that I don't have any experience with the CMake boost-bzip2 detection. Assuming that works and produces all of the right compiler flags, this seems like it should be fine.

arahlin:
CMake doesn't seem to be able to find the bzip2 component on any machine I've tried (OSX, the github action Ubuntu machine, amundsen), and instead gives the error `Could NOT find boost_bzip2 (missing: boost_bzip2_DIR)`.  If I manually set the `Boost_BZIP2_FOUND` variable to TRUE, regardless of whether cmake is able to find the component, then I end up with a working reader/writer.

cnweaver:
Yeah, now that I look at it, it seems that for IceTray we not only don't use the boost component to search for this; we've never actually done anything to detect bzip2, and I guess just explode in obscure ways if it isn't present. E.g. we just seem to pull in the header and run with it unconditionally: [https://github.com/icecube/icetray-public/blob/main/icetray/private/open/open.cxx#L36](https://github.com/icecube/icetray-public/blob/main/icetray/private/open/open.cxx#L36).

arahlin:
> Yeah, now that I look at it, it seems that for IceTray we not only don't use the boost component to search for this; we've never actually done anything to detect bzip2, and I guess just explode in obscure ways if it isn't present. E.g. we just seem to pull in the header and run with it unconditionally: https://github.com/icecube/icetray-public/blob/main/icetray/private/open/open.cxx#L36.

Yeah, that's the problem I'm running into.  I think the right solution is to look for the bzip2 library directly, rather than the boost component.  @marius311 would you be willing to test that this works?

arahlin:
Ok, so this change doesn't actually resolve @marius311's problem, which is that he's got a version of boost built without bzip2, even though bzip2 is installed on the system.  One option is to add a configurable argument to manually disable bzip2 support.  Another (which I don't know how to do) is to somehow detect whether `libboost_iostreams` is built against bzip2 or not.

cnweaver:
It would be *possible* to inpect `libboost_iostreams` with `ldd`, `otool -L`, etc. but it might be fragile, and seems like a lot of work for a case I've never heard of happening before. Is it known if the peculiar boost build came from some package manager (and so other users are likely to run into it), or can it be written off as a problem for a single user, who will have to resolve it?

arahlin:
OK, I've added a manual override option.  Just add `-DWITH_BZIP2=FALSE` to your cmake call and it should not look for bzip2 nor try to compile boost bzip2 objects.  @marius311 one last test?

marius311 (approved):
Tested and works for me :pray: 

@cnweaver if you're curious the boost w/o bzip2 is a binary library coming from Julia's package manager

PR #114, opened 2023-08-17
fix pip/poetry for submodules, better pep440

marius311:
Couple small things that I only caught after https://github.com/CMB-S4/spt3g_software/pull/97 got merged into the private repo and I used in action. 

1. The private repo has tags like "autoprocessing-3.3" so the version string created wasn't PEP440 compliant (unlike this public repo which just has number tags). The solution in this PR is to strip out starting characters. Open to other suggestions.

2. `-e` instead of `-d` works if the repo is a submodule, in which case `.git` is a file not directory.

### DIFF
--- a/cmake/getvers.sh
+++ b/cmake/getvers.sh
@@ -6,9 +6,10 @@ set -e
 cd $1
 
 # PEP440-compliant version number for pyproject.toml
-if [ -d .git ]; then
-	# replaces first - with .dev and second - with +, so e.g. 0.3-154-gd36baf4a becomes 0.3.dev154+gd36baf4a
-	fullversion_pep440=$(echo $(git describe --always --tags 2>/dev/null) | sed 's/-/.dev/' | sed 's/-/+/')
+if [ -e .git ]; then
+	# removes starting characters, so e.g. autoprocessing-0.3-154-gd36baf4a becomes just 0.3-154-gd36baf4a
+	# then replaces first - with .dev and second - with +, so e.g. 0.3-154-gd36baf4a becomes 0.3.dev154+gd36baf4a
+	fullversion_pep440=$(echo $(git describe --always --tags 2>/dev/null) | sed 's/^[^0-9]*//' | sed 's/-/.dev/' | sed 's/-/+/')
 fi
 fullversion_pep440="${fullversion_pep440:-0.1.0+unknown}" # fallback for SVN or error above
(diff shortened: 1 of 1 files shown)


arahlin (approved):

arahlin:
> Is "autoprocessing-0.3" actually equivalent to "somethingelse-0.3"?

No, it's not.  But we do also store the full tag name, and the autoprocessing version actually gets recorded as a separate variable altogether, so it doesn't really matter for this purpose.

marius311:
So what if anything do I need to do?

nwhitehorn:
So here is my proposed solution:
- We don't commit this patch. I don't think it makes sense to declare the software version to be whatever numbers happen to be contained in the most recent git tag.
- We pick a new format for the tags. I would propose we use the local-version-identifier bits in PEP440 to make the SPT-private version strings be <public version>+<informative string>. For example, to make "autoprocessing-v3.3" be "0.3+autoprocessing-3.3", which is almost PEP-440 compliant. This code could replace hyphens with dots to make it fully PEP-440 compliant. Or we could do that in the tag.
- We should probably make a new release of the public repo, which hasn't had one since October 2020.

arahlin:
> So here is my proposed solution:
> 
> * We don't commit this patch. I don't think it makes sense to declare the software version to be whatever numbers happen to be contained in the most recent git tag.
> * We pick a new format for the tags. I would propose we use the local-version-identifier bits in PEP440 to make the SPT-private version strings be +. For example, to make "autoprocessing-v3.3" be "0.3+autoprocessing-3.3", which is almost PEP-440 compliant. This code could replace hyphens with dots to make it fully PEP-440 compliant. Or we could do that in the tag.
> * We should probably make a new release of the public repo, which hasn't had one since October 2020.

This seems like a reasonable solution to me.  If we're going that route, I'd prefer that we make PEP440-compliant tags rather than write code to make them so.

I think we still need the `-e` change for detecting the presence of the `.git` file/directory, so that part of this PR should remain.

arahlin:
@marius311 are you ok with the plan discussed above?  If so, I'll close this PR.

marius311:
That's fine by me. Anything else you need from me or you guys will handle the remaining work?

arahlin:
> That's fine by me. Anything else you need from me or you guys will handle the remaining work?

I think the rest of the work is on us.  Thanks!

Issue #113, opened 2023-08-16
Inconsistent numpy detection


arahlin:
I'm finding that numpy is not detected when cmake is run a second time after the initial setup.  This happens on my mac, but I don't think the issue is OS-specific.

The first time cmake is run I get this:

```
$ cmake ..
...
-- Found Python: /usr/local/Frameworks/Python.framework/Versions/3.11/bin/python3.11 (found version "3.11.4") found components: Interpreter Development Development.Module Development.Embed
-- Found NumPy: /usr/local/lib/python3.11/site-packages/numpy/core/include
-- Found SciPy: FOUND
-- Python modules will be installed to /usr/local/lib/python3.11/site-packages
...
```

The second time I get this:

```
$ cmake ..
-- Found NumPy: NOT FOUND
-- Found SciPy: FOUND
-- Python modules will be installed to /usr/local/lib/python3.11/site-packages
...
```

arahlin:
Looks like something funky with cached variables.  If I make the following change:

```
diff --git a/cmake/Spt3gBoostPython.cmake b/cmake/Spt3gBoostPython.cmake
index 8cc1bcd..fe9a0a7 100644
--- a/cmake/Spt3gBoostPython.cmake
+++ b/cmake/Spt3gBoostPython.cmake
@@ -16,8 +16,10 @@ execute_process(COMMAND ${Python_EXECUTABLE} -c "import numpy"
        RESULT_VARIABLE NUMPY_FOUND ERROR_QUIET)
 if(NUMPY_FOUND EQUAL 0)
        set(Python_NumPy_FOUND TRUE CACHE BOOL "Numpy found successfully")
+ message(STATUS "Found NumPy: FOUND")
 else(NUMPY_FOUND EQUAL 0)
        set(Python_NumPy_FOUND FALSE CACHE BOOL "Numpy found successfully")
+ message(STATUS "Found NumPy: NOT FOUND")
 endif(NUMPY_FOUND EQUAL 0)

 if(Python_NumPy_FOUND)
```

I end up with this output: 

```
$ cmake ..
-- Found NumPy: FOUND
-- Found NumPy: NOT FOUND
-- Found SciPy: FOUND
-- Python modules will be installed to /usr/local/lib/python3.11/site-packages
```

So numpy is found correctly (in that the import command works), but the cached variable isn't being read properly somehow.  Probably we're just doing something wrong with how the numpy (and scipy) detection works.

PR #115, opened 2023-08-17
Force caching of Python_NumPy_FOUND to fix repeated application of cmake

cnweaver:
This seems to fix https://github.com/CMB-S4/spt3g_software/issues/113 .


arahlin (approved):

arahlin:
I think longer-term we probably don't want to search for numpy / scipy every time cmake is run, and instead handle these the same way that other packages are handled, but this fix will work for the time being.

PR #116, opened 2023-08-30
HitsBinner module to compute simple hits maps

arahlin:
This module has the same functionality as the `MapBinner`, but emits map frames that contain only a single `H` (hits) map.

### DIFF
--- a/maps/src/MapBinner.cxx
+++ b/maps/src/MapBinner.cxx
@@ -18,8 +18,7 @@ public:
 	MapBinner(std::string output_map_id, const G3SkyMap &stub_map,
 	    std::string pointing, std::string timestreams,
 	    std::string detector_weights, std::string bolo_properties_name,
-	    bool store_stokes_map, bool store_weight_map,
-	    boost::python::object map_per_scan);
+	    bool store_weight_map, boost::python::object map_per_scan);
 	virtual ~MapBinner() {}
 
 	void Process(G3FramePtr frame, std::deque<G3FramePtr> &out);
@@ -37,10 +36,9 @@ private:
 	int map_per_scan_;
 	boost::python::object map_per_scan_callback_;
 
-	bool pol_;
 	bool units_set_;
 
-	G3SkyMapPtr stub_, T_, Q_, U_;
+	G3SkyMapPtr T_, Q_, U_;
 	G3SkyMapWeightsPtr map_weights_;
 	G3Time start_, stop_;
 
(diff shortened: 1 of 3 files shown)


maps/src/MapBinner.cxx
```diff
@@ -18,7 +18,8 @@ class MapBinner : public G3Module {
 	MapBinner(std::string output_map_id, const G3SkyMap &stub_map,
 	    std::string pointing, std::string timestreams,
 	    std::string detector_weights, std::string bolo_properties_name,
-	    bool store_weight_map, boost::python::object map_per_scan);
+	    bool store_stokes_map, bool store_weight_map,
```
nwhitehorn, line 21: I guess T is a stokes parameter, but it still seems a little odd as a name. Elsewhere we refer to the "map" and the "weight map". Can this be "store_map"?

maps/src/MapBinner.cxx
```diff
@@ -206,9 +220,8 @@ MapBinner::Process(G3FramePtr frame, std::deque<G3FramePtr> &out)
 		}
 
 		if (map_weights_) {
-			out_frame->Put((Q_) ? "Wpol" : "Wunpol", map_weights_);
-			map_weights_ = G3SkyMapWeightsPtr(new G3SkyMapWeights(
-			    T_, T_->GetPolConv() != G3SkyMap::ConvNone));
+			out_frame->Put(pol_ ? "Wpol" : "Wunpol", map_weights_);
```
nwhitehorn, line 223: Why not just clone the old `map_weights_` here like the T code does? Then you don't have to maintain any parallel state.

maps/src/MapBinner.cxx
```diff
@@ -260,14 +273,16 @@ MapBinner::Process(G3FramePtr frame, std::deque<G3FramePtr> &out)
...
+		if (T_)
+			T_->units = timestreams->GetUnits();
 		if (Q_)
 			Q_->units = timestreams->GetUnits();
 		if (U_)
 			U_->units = timestreams->GetUnits();
 		units_set_ = true;
 	} else {
-		if (T_->units != timestreams->GetUnits())
+		if (stub_->units != timestreams->GetUnits())
```
nwhitehorn, line 285: It seems fine to just not do this check if T_ isn't set. There also isn't any reason the weights-only case even needs to have timestreams (and it might actually be convenient if it did not require them for checking coverage on sim stubs)

maps/src/MapBinner.cxx
```diff
@@ -394,15 +411,19 @@ MapBinner::BinTimestream(const G3Timestream &det, double weight,
...
 		for (size_t i = 0; i < det.size(); i++) {
-			(*T)[detpointing[i]] += pcoupling.t*det[i];
-			(*Q)[detpointing[i]] += pcoupling.q*det[i];
-			(*U)[detpointing[i]] += pcoupling.u*det[i];
+			if (T)
+				(*T)[detpointing[i]] += pcoupling.t*det[i];
+			if (Q)
+				(*Q)[detpointing[i]] += pcoupling.q*det[i];
+			if (U)
+				(*U)[detpointing[i]] += pcoupling.u*det[i];
```
nwhitehorn, line 419: Have you benchmarked this? Adding these branches could cause meaningful pessimization.

nwhitehorn (commented):
This is a pretty invasive patch, and I wonder if there are easier ways to implement it that don't pessimize the normal operating case. For example, I suspect that running with store_stokes_map = False only uses a little less RAM than when store_stokes_map = True and is only very marginally faster, if at all; couldn't you implement this patch more simply by simply not storing the real map at the end? I guess I'm a little unclear on the use case here.

nwhitehorn (approved):
Sweet

PR #117, opened 2023-09-05
Improvements to maps.CoaddMaps

arahlin:
This PR prepares for cleaning up the coadding tools in the private SPT-3G repo by migrating some features from there to the public `maps.CoaddMaps` utility.  In particular:

- Store the `.coadd_frame`  or `.coadd_frames` attribute persistently to enable extracting the coadded frames from the pipeline on completion.
- Allow collating input frames by map Id into separate coadds.
- Allow ignoring the weight state of input maps.  By default, the coadd will be made by ensuring that all input maps have weights applied.
- Track the map Ids and input files that go into building each coadd.
- A command-line utility for coadding maps directly from a set of input files into a single output file.


nwhitehorn (approved):
Looks good to me, but depends on another PR, so please (obviously) hold off merging.

PR #118, opened 2023-09-06
Track source filename for G3Frame objects that are read in using G3Reader

arahlin:
Add a ``G3Frame._filename`` attribute to frames, set when read in with G3Reader (or ARCFileReader).  This simplifies keeping track of input filenames further down the pipeline.  Documentation has been added to clarify that the attribute is not serialized, and therefore does not survive pickling or passing around in a multiprocessing situation.


nwhitehorn (changes requested):
I would really like to have this interface be more private -- initial underscores at minimum. It is very fragile and only for debugging.

nwhitehorn (approved):
Looks great -- thanks!

PR #126, opened 2023-10-24
StokesVector and MuellerMatrix Improvements

arahlin:
* Add StokesVector constructor from polarization angle and efficiency
* Add MuellerMatrix constructor from StokesVector components
* Add `MuellerMatrix.from_vector` method for populating an existing MuellerMatrix from a StokesVector
* Add `rotate_pol` methods for both structures (for use eventually in handling boresight rotation)
* Use these features in MapBinner and MapMockObserver


nwhitehorn:
Why the case change in det() and inv()? Any expected fallout? I don't think those are used much.

arahlin:
> Why the case change in det() and inv()? Any expected fallout? I don't think those are used much.

These are relatively minimal structures, so I thought their methods should be lowercase to indicate that in some way.  Could revert to camel case if you prefer.

These structures are only really used within the maps library AFAIK, so I don't expect any issues with the case change.

nwhitehorn:
If the usage is fully internal, I don't think it matters either way.

nwhitehorn (approved):
Looks great! This has been needed for a long time.

PR #130, opened 2023-11-19
Add new pointing model adjustment parameters

arahlin:
Add new `OnlinePointingModelCorrection` mapping to include commanded adjustments to tilts, flexure and collimation parameters.  This will be used by the machine learning pointing model correction code for EHT.


pchich:
I think you got everything, looks good!

PR #133, opened 2024-01-10
Smarter handling of map polarization properties

arahlin:
Add TT, TQ, etc to the `MapPolType` enumerator, and set the `pol_type` attribute for each map weights item.  Add a new `SetPolConv` pipeline module for changing polarization convention.  Additionally, the TU and QU attributes of a `G3SkyMapWeights` object are multiplied by -1 (along with the U map) if their polarization convention is changed from IAU to COSMO or vice versa using the `SetPolConv` module.  The `G3SkyMap.pol_conv` attribute is now a simple public attribute, and will not result in changing the sign of U/TU/QU maps when updated directly.

Where possible, determine parameters from the input map stub, similarly to the way these are handled in the `MapBinner` pipeline module.  Most notably, the `G3SkyMapWeights` constructor now determines its polarization state based on the `pol_conv` property of the input reference map.  Similarly, the `InjectMapStub` pipeline module determines whether the output frame is weighted and/or polarized based on the properties of the input stub.

### DIFF
--- a/maps/README.rst
+++ b/maps/README.rst
@@ -27,7 +27,7 @@ The following attributes are common to all G3SkyMap subclasses:
   The Stokes polarization of the map object, which is an instance of the ``MapPolType`` enum, and can have the value ``T``, ``Q``, ``U`` or None.
 
 ``pol_conv``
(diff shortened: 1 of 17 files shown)


cailmdaley (commented):
This looks great! My only suggestion would be to consider adding a module to consistently modify `pol_conv` for a weighted map frame, since it's no longer as simple as setting `U.pol_conv` (which I assume is what users will try to do at first).

nwhitehorn (approved):
Looks great, please merge

PR #136, opened 2024-01-22
Make FlattenPol set flat_pol attribute of T maps and weights

cailmdaley:
`FlattenPol` sets the `flat_pol` attribute of QU maps and weights to keep track of whether a frame has been flattened, but does not set this attribute for T maps and weights. While `FlattenPol` does not affect T, a frame's T map is often used as a "stub" containing metadata for the generation of subsequent data products. As a result, polarized data products generated from a T stub will not inherit the `flat_pol` attribute as one might expect. After discussion with @arahlin, we decided to make `FlattenPol` modify `flat_pol` for T as well as Q and U.

### DIFF
--- a/maps/python/map_modules.py
+++ b/maps/python/map_modules.py
@@ -96,9 +96,7 @@ def ApplyWeights(frame):
     frame["T"] = tmap
     if "Wpol" in frame:
         frame["Q"] = qmap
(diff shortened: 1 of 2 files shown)


arahlin (approved):
Looks good, thanks!

PR #137, opened 2024-01-24
Skip empty files in G3Reader

### LINKED ISSUE #134: G3Reader fails on multiple files when last file is empty
Create empty file:
```
In [3]: wr = core.G3Writer('empty.g3')

In [4]: del wr
```
Create non-empty file:
```
In [5]: fr = core.G3Frame()

In [6]: fr['A'] = 0

In [7]: wr = core.G3Writer('full.g3')

In [8]: wr(fr)
Out[8]: [<spt3g.core.G3Frame at 0x7fd5034ac450>]

In [9]: del wr
```
Read empty file only:
```
In [10]: p = core.G3Pipeline()

In [11]: p.Add(core.G3Reader, filename = ['empty.g3'])

In [12]: p.Run()  #runs fine, no errors
```
Read full file first and then empty file:
```
In [13]: p = core.G3Pipeline()

In [14]: p.Add(core.G3Reader, filename = ['full.g3', 'empty.g3'])

In [15]: p.Run()
ERROR (G3Reader) 18-Jan-2024:11:11:52 CST: Exception raised while reading file empty.g3 (G3Reader.cxx:98 in virtual void G3Reader::Process(G3FramePtr, std::deque<boost::shared_ptr<G3Frame> >&))
WARN (Unknown) 18-Jan-2024:11:11:52 CST: Exception in module "spt3g.core.G3Reader" (N6cereal9ExceptionE): Failed to read 1 bytes from input stream! Read 0 (G3Pipeline.cxx:119 in size_t {anonymous}::PushFrameThroughQueue(G3FramePtr, bool, bool, rusage&, std::vector<{anonymous}::G3Pipeline_mod_data>&, std::vector<{anonymous}::G3Pipeline_mod_data>::iterator, int&, std::deque<{anonymous}::G3Pipeline_proc_data>&, G3FramePtr&))
RuntimeError                              Traceback (most recent call last)
<ipython-input-15-ecb398338890> in <module>
----> 1 p.Run()

RuntimeError: Failed to read 1 bytes from input stream! Read 0

```

Note: other combinations (read full file only, read empty file first and full file second) also work fine.

arahlin:
Avoid throwing an IO error when an empty G3 file is included in the input filenames by moving on to the next file in the list, until reaching a non-empty file or the end of the list.  Issues an error message for every empty file that is encountered.

Fixes #134.

### DIFF
--- a/core/include/core/G3Reader.h
+++ b/core/include/core/G3Reader.h
@@ -26,6 +26,7 @@ private:
 	boost::iostreams::filtering_istream stream_;
 	int n_frames_to_read_;
 	int n_frames_read_;
+	int n_frames_cur_;
 	float timeout_;
 	bool track_filename_;
 
--- a/core/src/G3Reader.cxx
+++ b/core/src/G3Reader.cxx
@@ -7,7 +7,8 @@
 G3Reader::G3Reader(std::string filename, int n_frames_to_read,
     float timeout, bool track_filename) :
     prefix_file_(false), n_frames_to_read_(n_frames_to_read),
-    n_frames_read_(0), timeout_(timeout), track_filename_(track_filename)
+    n_frames_read_(0), n_frames_cur_(0), timeout_(timeout),
+    track_filename_(track_filename)
 {
 	boost::filesystem::path fpath(filename);
 	if (filename.find("://") == std::string::npos &&
@@ -20,7 +21,8 @@ G3Reader::G3Reader(std::string filename, int n_frames_to_read,
 G3Reader::G3Reader(std::vector<std::string> filename, int n_frames_to_read,
     float timeout, bool track_filename) :
     prefix_file_(false), n_frames_to_read_(n_frames_to_read),
-    n_frames_read_(0), timeout_(timeout), track_filename_(track_filename)
+    n_frames_read_(0), n_frames_cur_(0), timeout_(timeout),
+    track_filename_(track_filename)
 {
 	if (filename.size() == 0)
 		log_fatal("Empty file list provided to G3Reader");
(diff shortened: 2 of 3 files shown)


nwhitehorn:
Per the comment on the issue, this was a deliberate choice as a way to fail loudly on failed cluster jobs. Can we hide the new behavior behind an option so that people who turn it on are aware of consequences?

arahlin:
> Per the comment on the issue, this was a deliberate choice as a way to fail loudly on failed cluster jobs. Can we hide the new behavior behind an option so that people who turn it on are aware of consequences?

Sure, that's probably a good idea.  @samtomguns are you ok with this solution?

arahlin:
> > Per the comment on the issue, this was a deliberate choice as a way to fail loudly on failed cluster jobs. Can we hide the new behavior behind an option so that people who turn it on are aware of consequences?
> 
> Sure, that's probably a good idea. @samtomguns are you ok with this solution?

I'll add that the current failure mode is inconsistent -- if the only input file is empty, or just the first file in the list is empty, then the code runs without failure.

samtomguns:
is something special about the last file from the perspective of cluster jobs? wouldn't it make more sense to have it fail on every empty file unless some extra option is turned off?

arahlin:
I think the right solution is to check whether a file emitted any frames, and raise an error if it didn't (or raise a warning and skip to the next file, if the user sets that option)

nwhitehorn:
This already fails on the presence of any empty file. We can change the behavior, of course -- I just wanted everyone to be on the same page about why the current behavior exists before we do.

nwhitehorn:
Sorry, that was wrong. That was what it *should* have done. Since it isn't, let's just merge the PR.

nwhitehorn (approved):
Per comment on issue, what this was supposed to handle was broken to begin with, so this is the closest to existing behavior as it actually is.

arahlin:
Sounds good.  As a compromise, I've added an error message for when the reader encounters an empty file.

PR #138, opened 2024-01-24
Disable FITS image compression by default

### LINKED ISSUE #135: GZIP_2 compression of a flat-sky map seems lossy in some cases
When we save a flat-sky map in a FITS file via `maps.fitsio.save_skymap_fits`, the default option for the argument `compress` is `'GZIP_2'`. This compression scheme is supposed to be lossless, but that does not seem to be the case when I tested the compression scheme on an SPT-3G winter field coadd.

Here is an example:
```
In[4]: frame = core.G3File(path_to_a_coadd).next()


# Default save

In [5]: maps.fitsio.save_skymap_fits(output_path, frame["T"], Q=frame["Q"], U=frame["U"])

In [6]: frame_debug = maps.fitsio.load_skymap_fits(output_path)

In [7]: numpy.allclose(frame["T"], frame_debug["T"])
Out[7]: False


# No-compression save

In [18]: maps.fitsio.save_skymap_fits(output_path, frame["T"], Q=frame["Q"], U=frame["U"], compress=False, overwrite=True)

In [19]: frame_debug = maps.fitsio.load_skymap_fits(output_path)

In [20]: numpy.allclose(frame["T"], frame_debug["T"])
Out[20]: True
```

arahlin:
Compression of floating point data using the GZIP_2 fits compression algorithm is *not* lossless.  To avoid pitfalls with this algorithm, compression is disabled by default.  Also expose the `quantize_level` keyword argument to enable adjustment of the compression algorithm, if compression is desired.

See the astropy FITS image documentation for further information: https://docs.astropy.org/en/stable/io/fits/api/images.html

Closes #135.

### DIFF
--- a/maps/python/fitsio.py
+++ b/maps/python/fitsio.py
@@ -481,7 +481,7 @@ def parse_wcs_header(header):
 
 @core.usefulfunc
 def save_skymap_fits(filename, T, Q=None, U=None, W=None, overwrite=False,
-                     compress='GZIP_2', hdr=None):
+                     compress=False, quantize_level=16.0, hdr=None):
     """
     Save G3 map objects to a fits file.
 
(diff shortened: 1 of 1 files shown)


Wei-Q:
Thank you for making these changes, and they look great to me!

PR #140, opened 2024-02-06
Efficient storage of integer G3Map and G3Vector objects

### LINKED ISSUE #122: G3MapVectorInt and G3MapInt should use int64
Currently these use int32, which is inconsistent with G3VectorInt.

arahlin:
Use `int64_t` for in-memory representation of integer values, but store as 8-, 16-, 32- or 64-bit integers on disk, depending on the bit depth of the underlying data.  Use a consistent ABI for `G3Vector` and `G3Map` serialization, and ensure backwards compatibility for v1 `int32_t` objects on disk.

Closes #122.

### DIFF
--- a/core/src/G3Map.cxx
+++ b/core/src/G3Map.cxx
@@ -14,7 +14,7 @@ void G3Map<std::string, int64_t>::load(A &ar, const unsigned v)
 	G3_CHECK_VERSION(v);
 
 	ar & cereal::make_nvp("G3FrameObject",
-			      cereal::base_class<G3FrameObject>(this));
+	    cereal::base_class<G3FrameObject>(this));
 	int store_bits = 32;
 	if (v >= 2)
 		ar & cereal::make_nvp("store_bits", store_bits);
@@ -22,7 +22,7 @@ void G3Map<std::string, int64_t>::load(A &ar, const unsigned v)
 	switch(store_bits) {
 	case 64:
 		ar & cereal::make_nvp("map",
-				      cereal::base_class<std::map<std::string, int64_t> >(this));
+		    cereal::base_class<std::map<std::string, int64_t> >(this));
 		break;
 	case 32:
 		load_as<A, int32_t>(ar, *this);
@@ -42,7 +42,7 @@ void G3Map<std::string, int64_t>::save(A &ar, const unsigned v) const
 {
 	// v == 2
 	ar & cereal::make_nvp("G3FrameObject",
-			      cereal::base_class<G3FrameObject>(this));
+	    cereal::base_class<G3FrameObject>(this));
 	// Count the interesting bits, and convert to nearest power of 2.
 	int sig_bits = bit_count(*this);
 	int store_bits = 8;
@@ -61,7 +61,7 @@ void G3Map<std::string, int64_t>::save(A &ar, const unsigned v) const
 		break;
 	default:
 		ar & cereal::make_nvp("map",
-				      cereal::base_class<std::map<std::string, int64_t> >(this));
+		    cereal::base_class<std::map<std::string, int64_t> >(this));
 	}
 }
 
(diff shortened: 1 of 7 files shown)


arahlin:
Tagging @mhasself here, in case you have opinions on this implementation, since you wrote the G3VectorInt implementation this is copied from.

core/src/G3Map.cxx
```diff
@@ -4,6 +4,201 @@
...
+		save_as<A, int64_t, int32_t>(ar, *this);
+		break;
+	default:
+		ar & cereal::make_nvp("map",
+				      cereal::base_class<std::map<std::string, int64_t> >(this));
+	}
+}
+
+static
+int bit_count_vector(std::map<std::string, std::vector<int64_t> > const &d) {
```
mhasself, line 102: Really this is `bit_count_map_vector`, right?

In the case of G3MapVectorInt, do you think it's efficient enough, to compute a single value of store_bits for *all* the data in all the entries?  I would have thought that in many cases you'd have heterogeneous data, from one field to the next, and thus benefit from computing store_bits separately for each entry.

core/src/G3Map.cxx
```diff
@@ -4,6 +4,201 @@
 #include <container_pybindings.h>
 #include <serialization.h>
 
+/* Special load/save for int64_t. */
```
mhasself, line 7: Maybe around mention that a similar encoding is also used in G3Vector.cxx?

These bit_count functions (incl. the one in g3Vector.cxx) could be templated and put in a header file pretty easily to reduce duplication.  Take advantage of bit_count([a, b, c]) = max([bit_count(a), bit_count(b), bit_count(c)]).

mhasself (approved):
Thanks for the tag!  Looks good to me.  There's a lot of repetitive code now ... but if the alternative is template mazes then that's not always better.

Couple suggestions in line, but don't stop this train for me.

core/src/G3Map.cxx
arahlin: Good call, I've rearranged G3MapVectorInt to compress each vector independently.

core/src/G3Map.cxx
arahlin: Moved the storage functions to separate (private) header/source files, and used templates where appropriate.

nwhitehorn (commented):
Some minor code comments in line. How much time does bit_count() take? Is there any meaningful run time impact from this change?

The code is a little spaghetti-ish, but I don't see a viable way to avoid that.

core/src/G3Map.cxx
```diff
@@ -3,6 +3,152 @@
...
+	ar & cereal::make_nvp("len", len);
+
+	for (auto const &i: *this) {
+		ar & cereal::make_nvp("key", i.first);
+
+		// Count the interesting bits, and convert to nearest power of 2.
+		int sig_bits = bit_count(i.second);
+		int store_bits = 8;
+		while (store_bits < sig_bits)
+			store_bits *= 2;
```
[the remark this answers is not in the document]
nwhitehorn: Apparently __builtin_align_up() is a clang-ism that the GCC people haven't brought in yet. There's also this: https://www.boost.org/doc/libs/1_68_0/boost/align/align_up.hpp

But the while loop is fine.

arahlin:
I haven't profiled this -- @mhasself do you know how efficient the `bit_count` function is?

nwhitehorn (approved):

mhasself:
Against my better schedule management ... here is a (relative) profiling.  The bit_count function takes 10 ns / sample on my laptop.  Tests done with 1k to 1M elements.  It seems to take about 10% more time than this simple accumulation function:
```
int not_bit_count_v2(const std::vector<int64_t> &d) {
    int sum = 0;
    for (auto c: d)
        sum += c;
    return sum;
}
```

So I think it's close enough to maxed out.

arahlin:
Cool, good to know!

PR #141, opened 2024-02-09
Avoid unnecessary error log messages in MapBinner

arahlin:
When multiple MapBinners are used in a single pipeline with temporally-split timestreams (e.g. separate maps for left and right scans), each binner emits an error message when it doesn't find the appropriate timestreams in every scan frame.  This PR is intended to avoid these unnecessary error messages, and instead issue a single error message if an empty map is emitted at the end of the pipeline.


nwhitehorn (approved):
I don't love that this gives no indication of a problem until (potentially) very late. Can you add a lower-level (log_info or something) message that is suppressed by default for each frame?

PR #144, opened 2024-02-27
Testing environment cleanup

arahlin:
Set path environment variables for tests from the build directory.  Ensures that the correct libraries are always used, without having to make changes to the parent shell environment.


nwhitehorn (approved):
Looks good -- thanks for turning over some very old stones.

PR #146, opened 2024-02-28
G3PythonContext class for handling the Python GIL

### LINKED ISSUE #145: Consolidate python GIL / threads context handling
See potential implementation in #69, which can be split off into a separate PR.

arahlin:
This PR creates a new class that simplifies initialization of python threads, as well as acquiriing / releasing the Python global interpreter lock in various contexts.

Use cases include:

1. Ensuring that Py_Initialize() is properly called at the beginning of a program that is expected to interact with the python interpreter, and also that Py_Finalize() is called when the program is finished.
2. Ensuring that the current thread state is saved and the GIL released as necessary, e.g. for IO operations, and then the thread state is restored on completion.
3. Ensuring that the GIL is acquired for one-off interaction with the python interpreter, and released when complete.

A G3PythonContext object is used throughout the library code for cases 2 and 3. Case 1 is handled by a G3PythonInterpreter object constructed at the main program/thread level.  If the python interpreter has not been initialized (i.e. the compiled program is expected to be purely in C++), then the library context objects are essentially no-op. If the python interpreter is initialized (e.g. inside a python program or command-line interface), then these library context objects will handle the GIL appropriately.

See the examples/cppexample.cxx C++ program for a simple implementation of the above behavior in a compiled program.

Closes #145.

### DIFF
--- a/core/include/core/G3PipelineInfo.h
+++ b/core/include/core/G3PipelineInfo.h
@@ -2,15 +2,15 @@
 #define _G3_PIPELINEINFO_H
 
 #include <G3Frame.h>
-#include <G3Map.h>
 #include <vector>
+#include <map>
 
 class G3ModuleConfig : public G3FrameObject {
 public:
 	std::string modname;
 	std::string instancename;
 
-	G3MapFrameObject config;
+	std::map<std::string, boost::python::object> config;
 
         template <class A> void load(A &ar, unsigned v);
         template <class A> void save(A &ar, unsigned v) const;
@@ -54,7 +54,7 @@ namespace cereal {
         template <class A> struct specialize<A, G3ModuleConfig, cereal::specialization::member_load_save> {};
 }
 
-G3_SERIALIZABLE(G3ModuleConfig, 2);
+G3_SERIALIZABLE(G3ModuleConfig, 1);
 G3_SERIALIZABLE(G3PipelineInfo, 2);
 
 #endif
--- a/core/include/core/G3PipelineInfo.h
+++ b/core/include/core/G3PipelineInfo.h
@@ -2,15 +2,15 @@
 #define _G3_PIPELINEINFO_H
 
 #include <G3Frame.h>
+#include <G3Map.h>
 #include <vector>
-#include <map>
 
 class G3ModuleConfig : public G3FrameObject {
 public:
 	std::string modname;
 	std::string instancename;
 
-	std::map<std::string, boost::python::object> config;
+	G3MapFrameObject config;
 
         template <class A> void load(A &ar, unsigned v);
         template <class A> void save(A &ar, unsigned v) const;
@@ -20,7 +20,6 @@ public:
 
 	bool operator ==(const G3ModuleConfig &) const;
 
-	~G3ModuleConfig();
 private:
 	SET_LOGGER("G3ModuleConfig");
 };
@@ -55,7 +54,7 @@ namespace cereal {
         template <class A> struct specialize<A, G3ModuleConfig, cereal::specialization::member_load_save> {};
 }
 
-G3_SERIALIZABLE(G3ModuleConfig, 1);
+G3_SERIALIZABLE(G3ModuleConfig, 2);
 G3_SERIALIZABLE(G3PipelineInfo, 2);
 
 #endif
--- a/core/include/core/G3PipelineInfo.h
+++ b/core/include/core/G3PipelineInfo.h
@@ -20,6 +20,7 @@ public:
 
 	bool operator ==(const G3ModuleConfig &) const;
 
+	~G3ModuleConfig();
 private:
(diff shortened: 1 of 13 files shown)


nwhitehorn:
How do you handle lock-order reversals during destructor calls (and GIL acquisition)? This is the rock we've always hit before: shared_ptr deleters can call into Python at unpredictable times and the only way to avoid deadlocks or crashes is to ensure that no shared pointers that Python has even potentially seen ever go out of scope with any non-GIL locks held and -- if any such pointers do go out of scope without the GIL held, Boost is patched to acquire the GIL in every shared pointer deleter when calling into Python. In practice, since it is hard to identify which pointers may have touched Python, this means the only places we have been able to use threading are those where all handled objects originate in C++ (i.e. the DfMux collector code).

cozzyd:
Assuming Sasha adapted my attempt at creating a similar RAII class in the json_output branch without adding any additional magic, I think unfortunately it doesn't solve that problem. 

I never got multithreaded code that can touch G3PipelineInfo  (which contains arbitrary python things) to never eventually deadlock, presumably as you said, because the shared pointer destructor can happen in an unpredictable context and could not find a good solution.

arahlin:
I basically don't serialize anything that isn't a frameobject in the G3PipelineInfo structure, and just bail onto the next thing.  It's not a great solution, and probably doesn't solve the deadlock issue, but it does allow the test program to at least read a PipelineInfo frame without segfaulting.

cozzyd:
I'll merge this branch into my json_output branch and see if the deadlock issues still happen or not :)

arahlin:
> I think the best course here is to either initialize the Python interpreter by hand in the C++ example or to detect that one doesn't exist and bail.

Yeah, that's the direction I've taken this implementation.  There's no reason to initialize the interpreter in library code.  I can add better words to that effect if that's not clear from the current description.

cozzyd:
Is there any case you would call this with anything other than (...,true, false) or (...,false,true)? 

If not, it may be a bit friendlier to separate initialization from the GIL context by introducing something like `G3PythonInitializer` instead for (...,false,true) case ?

arahlin:
> Is there any case you would call this with anything other than (...,true, false) or (...,false,true)?
> 
> If not, it may be a bit friendlier to separate initialization from the GIL context by introducing something like `G3PythonInitializer` instead for (...,false,true) case ?

I guess the way it's implemented, you can initialize the interpreter and either hold onto the GIL (true, true) or immediately release it (false, true).  I wouldn't be opposed to creating two (or three?) separate objects for this, though, if that makes things clearer.

cozzyd:
So I merged this into my json_output branch and found I still get segfaults when reading files with G3PipelineInfo in them concurrently. 

However, https://github.com/CMB-S4/spt3g_software/pull/69/commits/0eea5ee8a93c0dd48011f314f12dcff6dfdc125d seems to "fix" it (or at least make seg faults rare enough that I haven't come across them yet in hundreds of thousands of attempts with 8 threads). I can't say it's the only additional place we need to hold the GIL, but I think it makes sense that we need to hold the GIL here, if you want to cherry pick it...

arahlin:
> So I merged this into my json_output branch and found I still get segfaults when reading files with G3PipelineInfo in them concurrently.
> 
> However, [0eea5ee](https://github.com/CMB-S4/spt3g_software/commit/0eea5ee8a93c0dd48011f314f12dcff6dfdc125d) seems to "fix" it (or at least make seg faults rare enough that I haven't come across them yet in hundreds of thousands of attempts with 8 threads). I can't say it's the only additional place we need to hold the GIL, but I think it makes sense that we need to hold the GIL here, if you want to cherry pick it...

Have you tried running your PR without the interpeter initialized?

cozzyd:
The executable in my PR  did eventually crash without enabling the interpreter, but I think it was related to the "global" thing you just removed (the crash was in the boost::python::object destructor).  After your latest changes, it doesn't crash anymore in 500,000 trials with 32 threads... nicely done!

arahlin:
Great!  So I ended up just yanking all of the repr() machinery out of the C++ side of the module config code, and letting the python side handle creating and/or eval'ing the strings if necessary.  @nwhitehorn how does this look to you?

nwhitehorn:
This seems like a nice approach. Hats off! Would it be possible to split it into two pieces, one for the G3ModuleConfig stuff and one for the G3PythonContext tidying? They look cleanly separable if you do the G3PythonContext updates first.

arahlin:
> This seems like a nice approach. Hats off! Would it be possible to split it into two pieces, one for the G3ModuleConfig stuff and one for the G3PythonContext tidying? They look cleanly separable if you do the G3PythonContext updates first.

Sure, I'll move the ModuleConfig bits to a separate PR.

nwhitehorn (approved):
Looks good. Thanks!

PR #147, opened 2024-03-01
Isolate serialization of python objects in G3PipelineInfo interface

arahlin:
This PR adds logic throughout the G3PipelineInfo and G3ModuleConfig class definitions to enable them to serialize appropriately in a pure-C++ program.  Notably, any serialization of python objects is handled only on interface with the python interpreter, and otherwise stored in C++ memory with the string representation.


cozzyd:
This looks good to me!

nwhitehorn:
Just so I understand fully from reading the code: every object gets repr() called on it and the repr() of everything is now stored. For frame objects, the frame object is *also* serialized. On deserialization, any frame objects are instantiated, but things where just the repr() is stored are only evaluated lazily. Is that correct?

arahlin:
Yeah that's right.  I'm going to rearrange a bit so that the repr for frame objects in v1 files doesn't need to be grabbed on load, so that we don't need to grab the GIL in C++ only code ever.

nwhitehorn:
That sounds great. I'll hold off formally approving until those commits are in, but this is a *really* nice solution to the problem.

arahlin:
Should be all set now.  Thanks!

nwhitehorn (approved):
Looks great!

Issue #128, opened 2023-11-11
C++11 mode and -Werror prevent building with new Boost versions


cnweaver:
```
In file included from /Users/cweaver/Work/CMB-S4/spt3g_software-public-git/core/src/G3Vector.cxx:3:
In file included from /Users/cweaver/Work/CMB-S4/spt3g_software-public-git/core/include/core/container_pybindings.h:5:
In file included from /Users/cweaver/Work/CMB-S4/spt3g_software-public-git/core/include/core/G3Map.h:6:
In file included from /Users/cweaver/Work/CMB-S4/spt3g_software-public-git/core/include/core/G3Quat.h:6:
In file included from /usr/local/include/boost/math/quaternion.hpp:14:
/usr/local/include/boost/math/tools/config.hpp:23:6: error: "The minimum language standard to use Boost.Math will be C++14 starting in July 2023 (Boost 1.82 release)" [-Werror,-W#warnings]
#    warning "The minimum language standard to use Boost.Math will be C++14 starting in July 2023 (Boost 1.82 release)"
     ^
1 error generated.
```
The simplest solution is to change the language mode to C++14, but I would like to first verify that this will be available on all platforms we consider supported.

PR #148, opened 2024-03-01
Update language dialect to C++14

cnweaver:
Fixes https://github.com/CMB-S4/spt3g_software/issues/128.


arahlin (approved):

cozzyd:
Note that we should update the minimum supported compiler in the readme to whatever is necessary (GCC 5?)

cnweaver:
Yes, good point, it [looks like](https://en.cppreference.com/w/cpp/compiler_support/14) is would indeed be GCC >=5 or clang >=3.4.

PR #149, opened 2024-03-04
Update compiler requirements for C++14

cnweaver:
Follow-up to https://github.com/CMB-S4/spt3g_software/pull/148.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -19,7 +19,7 @@ This depends on Boost and cmake, as well as the usual Python packages. Some addi
 
 Minimum versions:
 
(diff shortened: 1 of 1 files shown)


arahlin:
> those versions seem correct to me, based on docs. This might be slightly painful for EL7 systems that have gcc 4.8 as a default, but it seems reasonable to use a standard 10 years after it is adopted...

EL7 is EOL this year, so this is probably ok. We [SPT] also use clustertools (https://github.com/SouthPoleTelescope/clustertools) with more recent compilers on any EL7 systems anyway.

PR #150, opened 2024-03-04
Update minimum cmake version to 3.5

arahlin:
The latest version of cmake (3.28.3) warns that versions < 3.5 will not be supported in the near future.


cozzyd:
update README.rst as well?

arahlin:
Bump.  @nwhitehorn or @cnweaver is this ok to merge?

nwhitehorn (approved):
Yes, this looks great. I don't think anyone is using 2015-era cmake any more.

PR #151, opened 2024-03-07
Interface for consistent bolometer band string formatting

arahlin:
This enables handling spectrometer channels with closely spaced frequencies, e.g. for mapmaking.


adamanderson (approved):
Looks good. I have no suggested changes.

PR #152, opened 2024-03-21
Update dfmux HousekeepingConsumer for mkids firmware

arahlin:
Add optional entries for RF-ICE properties in housekeeping dump.  Also use a different `ReadoutSystem` entry in the wiring frame (`"RF-ICE"` instead of `"ICE"`), for use later in converting timestreams into physical units.

Maintain backwards compatibility with 3G firmware throughout.


arahlin:
Tested on an RF-ICE board with firmware version 1.8.6.

adamanderson (approved):

PR #153, opened 2024-03-29
Do not allow seek after EOF in G3Reader

arahlin:
Upstream merge of simonsobs/so3g#172


mhasself (approved):
Apparently I should just be using this! :)

PR #155, opened 2024-04-25
Add a read-only current_file attribute for G3MultiFileWriter

arahlin:
This is useful for subclassing this writer for, e.g. writing files to the grid.


nwhitehorn (approved):
Thanks! This feature should have been in from the beginning.

PR #156, opened 2024-05-03
Optionally ignore wiring information in HousekeepingConsumer

arahlin:
This enables running the HousekeepingConsumer alongside a Wiring frame generated from a hardware map using the WiringMapInjector.

### DIFF
--- a/dfmux/python/Housekeeping.py
+++ b/dfmux/python/Housekeeping.py
@@ -79,9 +79,9 @@ class HousekeepingConsumer(object):
 
     def __call__(self, frame):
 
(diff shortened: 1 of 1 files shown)


nwhitehorn:
Looks good to me from a code point of view. Does hidfmux not provide an in-band hardware map interface? Having one for the 3G electronics was a huge reliability win.

arahlin:
Hidfmux has the same interfaces as pydfmux for this purpose.  This PR just allows constructing the wiring map from the hidfmux HWM while also getting housekeeping frames.

nwhitehorn (approved):
Looks great, thanks!

PR #158, opened 2024-08-12
G3Writer as a python context

arahlin:
This PR enables using G3Writer and G3MultiFileWriter as a python context object, so that the object can be created and cleanly closed using the python `with` statement.  For example:

    with core.G3Writer(filename="path/to/file.g3") as writer:
        writer(frame1)
        writer(frame2)

This will open the output file, write the two frames to it, then properly close the file by appending an EndProcessing frame.


core/python/fileio.py
```diff
@@ -0,0 +1,50 @@
...
+        return frames[0]
+
+    __next__ = next
+
+
+# writer context methods
+
+
+def writer_enter(self):
+    return self
```
nwhitehorn, line 39: Should these have initial underscores? They are now getting pulled into the root Python context because of the import * above.

nwhitehorn (commented):
Looks mostly good, but see the comment about import * inline. Thanks!

core/python/fileio.py
arahlin: I added an `__all__` definition at the top to avoid polluting the namespace.

nwhitehorn: Indeed you did. Sorry I didn't see it!

nwhitehorn (approved):

PR #160, opened 2024-08-23
Fix and reenable all compiler warnings

arahlin:
* Fix nearly all compiler warnings on latest GCC and Clang compilers
* Ignore several unfixable compiler warnings using targeted `pragma GCC diagnostic` blocks.
* Rearrange github workflow to use latest compilers, and for easier integration with private repo


arahlin:
Some version of this warning pops up quite often, and I'm not sure of the best way to fix it.

```
In file included from /usr/local/include/boost/python/override.hpp:13,
                 from /usr/local/include/boost/python/wrapper.hpp:8,
                 from /usr/local/include/boost/python/object/value_holder.hpp:15,
                 from /usr/local/include/boost/python/object/class_metadata.hpp:14,
                 from /usr/local/include/boost/python/class.hpp:23,
                 from /usr/local/include/boost/python.hpp:18,
                 from /Users/arahlin/code/spt/spt3g_software_public/core/include/core/pybindings.h:11,
                 from /Users/arahlin/code/spt/spt3g_software_public/core/src/python.cxx:1:
In member function 'boost::python::converter::extract_rvalue<T>::result_type boost::python::converter::extract_rvalue<T>::operator()() const [with T = bool]',
    inlined from 'static bool boost::python::indexing_suite<Container, DerivedPolicies, NoProxy, NoSlice, Data, Index, Key>::base_contains(Container&, PyObject*) [with Container = std::vector<bool, std::allocator<bool> >; DerivedPolicies = boost::python::detail::final_vector_derived_policies<std::vector<bool, std::allocator<bool> >, true>; bool NoProxy = true; bool NoSlice = false; Data = bool; Index = long unsigned int; Key = bool]' at /usr/local/include/boost/python/suite/indexing/indexing_suite.hpp:286:66,
    inlined from 'static bool boost::python::indexing_suite<Container, DerivedPolicies, NoProxy, NoSlice, Data, Index, Key>::base_contains(Container&, PyObject*) [with Container = std::vector<bool, std::allocator<bool> >; DerivedPolicies = boost::python::detail::final_vector_derived_policies<std::vector<bool, std::allocator<bool> >, true>; bool NoProxy = true; bool NoSlice = false; Data = bool; Index = long unsigned int; Key = bool]' at /usr/local/include/boost/python/suite/indexing/indexing_suite.hpp:273:9:
/usr/local/include/boost/python/extract.hpp:185:11: error: '*(bool*)((char*)&x + offsetof(boost::python::extract<bool>,boost::python::extract<bool>::<unnamed>.boost::python::converter::extract_rvalue<bool>::m_data.boost::python::converter::rvalue_from_python_data<bool>::<unnamed>.boost::python::converter::rvalue_from_python_storage<bool>::storage))' may be used uninitialized [-Werror=maybe-uninitialized]
  185 |           );
      |           ^
In file included from /usr/local/include/boost/python/suite/indexing/vector_indexing_suite.hpp:9,
                 from /Users/arahlin/code/spt/spt3g_software_public/core/include/core/pybindings.h:12:
/usr/local/include/boost/python/suite/indexing/indexing_suite.hpp: In static member function 'static bool boost::python::indexing_suite<Container, DerivedPolicies, NoProxy, NoSlice, Data, Index, Key>::base_contains(Container&, PyObject*) [with Container = std::vector<bool, std::allocator<bool> >; DerivedPolicies = boost::python::detail::final_vector_derived_policies<std::vector<bool, std::allocator<bool> >, true>; bool NoProxy = true; bool NoSlice = false; Data = bool; Index = long unsigned int; Key = bool]':
/usr/local/include/boost/python/suite/indexing/indexing_suite.hpp:284:30: note: 'x' declared here
  284 |                 extract<Key> x(key);
      |                              ^
```

core/include/core/G3NetworkSender.h
```diff
@@ -45,7 +45,7 @@ class G3NetworkSender : public G3Module {
 			std::lock_guard<std::mutex> lg(lock);
 			// Refuse to put anything into stopped queues, and if a maximum
 			// queue size is set and would be exceeded, drop low priority data
-			if (die || (max_size && low_priority && queue.size()>=max_size))
+			if (die || (max_size && low_priority && queue.size()>=(size_t)max_size))
```
cnweaver, line 48: Myb we should just change `max_size` to be `size_t` in the first place? E.g. there's no reason to allow it to be negative.

CMakeLists.txt
```diff
@@ -42,7 +42,7 @@ set(CMAKE_SHARED_LIBRARY_PREFIX "libspt3g-")
 
 # Raise errors on every warning by default
 # (use target-specific options to disable particular warnings)
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Werror")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall")
```
cnweaver, line 45: I am always in favor of avoiding `-Werror`, since it tends to act as a punishment for anyone who tries to use a new compiler. As long as warnings are generally kept under control it should stick out if patches set new ones off.

core/include/core/serialization.h
```diff
@@ -72,7 +72,7 @@ _g3_class_version(T *)
 }
 
 #define G3_CHECK_VERSION(v) \
-	if (v > _g3_class_version(this)) \
+	if ((int)v > _g3_class_version(this)) \
```
cnweaver, line 75: This also looks like a case where the original signed side should be unsigned; [Cereal versions are unsigned](http://uscilab.github.io/cereal/assets/doxygen/group__Utility.html#gabea833652252591a51f75cb9ce31f44a), specifically `uint32_t`, so `_g3_class_version` should probably just be returning that in the first place.

core/src/G3Timestream.cxx
```diff
@@ -515,12 +521,6 @@ G3Timestream G3Timestream::operator +(double r) const
 	return ret;
 }
 
-static
-G3Timestream operator +(double l, const G3Timestream &r)
-{
-	return r+l;
-}
-
```
cnweaver, line 523: Was there something wrong with these functions besides that they weren't accessible outside this translation unit? They seem generally useful, and could just have the `static` dropped and declarations added to the header file.

core/src/G3Timestream.cxx
```diff
@@ -529,12 +529,6 @@ G3Timestream G3Timestream::operator -(double r) const
 	return ret;
 }
 
-static
-G3Timestream operator -(double l, const G3Timestream &r)
-{
-	return (r-l)*(-1);
```
cnweaver, line 535: This particular implementation does seem a bit dubious; if this function is brought back I would change it to use an explicit loop like the division function so that it makes only one pass over the data.

cnweaver (approved):
A few comments and nitpicks, but this looks like useful cleanup overall.

cnweaver:
> Some version of this warning pops up quite often, and I'm not sure of the best way to fix it. This is with Boost 1.85 and gcc-14 on my Intel Mac laptop.
> 
> ```
> In file included from /usr/local/include/boost/python/override.hpp:13,
>                  from /usr/local/include/boost/python/wrapper.hpp:8,
>                  from /usr/local/include/boost/python/object/value_holder.hpp:15,
>                  from /usr/local/include/boost/python/object/class_metadata.hpp:14,
>                  from /usr/local/include/boost/python/class.hpp:23,
>                  from /usr/local/include/boost/python.hpp:18,
>                  from /Users/arahlin/code/spt/spt3g_software_public/core/include/core/pybindings.h:11,
>                  from /Users/arahlin/code/spt/spt3g_software_public/core/src/python.cxx:1:
> In member function 'boost::python::converter::extract_rvalue<T>::result_type boost::python::converter::extract_rvalue<T>::operator()() const [with T = bool]',
>     inlined from 'static bool boost::python::indexing_suite<Container, DerivedPolicies, NoProxy, NoSlice, Data, Index, Key>::base_contains(Container&, PyObject*) [with Container = std::vector<bool, std::allocator<bool> >; DerivedPolicies = boost::python::detail::final_vector_derived_policies<std::vector<bool, std::allocator<bool> >, true>; bool NoProxy = true; bool NoSlice = false; Data = bool; Index = long unsigned int; Key = bool]' at /usr/local/include/boost/python/suite/indexing/indexing_suite.hpp:286:66,
>     inlined from 'static bool boost::python::indexing_suite<Container, DerivedPolicies, NoProxy, NoSlice, Data, Index, Key>::base_contains(Container&, PyObject*) [with Container = std::vector<bool, std::allocator<bool> >; DerivedPolicies = boost::python::detail::final_vector_derived_policies<std::vector<bool, std::allocator<bool> >, true>; bool NoProxy = true; bool NoSlice = false; Data = bool; Index = long unsigned int; Key = bool]' at /usr/local/include/boost/python/suite/indexing/indexing_suite.hpp:273:9:
> /usr/local/include/boost/python/extract.hpp:185:11: error: '*(bool*)((char*)&x + offsetof(boost::python::extract<bool>,boost::python::extract<bool>::<unnamed>.boost::python::converter::extract_rvalue<bool>::m_data.boost::python::converter::rvalue_from_python_data<bool>::<unnamed>.boost::python::converter::rvalue_from_python_storage<bool>::storage))' may be used uninitialized [-Werror=maybe-uninitialized]
>   185 |           );
>       |           ^
> In file included from /usr/local/include/boost/python/suite/indexing/vector_indexing_suite.hpp:9,
>                  from /Users/arahlin/code/spt/spt3g_software_public/core/include/core/pybindings.h:12:
> /usr/local/include/boost/python/suite/indexing/indexing_suite.hpp: In static member function 'static bool boost::python::indexing_suite<Container, DerivedPolicies, NoProxy, NoSlice, Data, Index, Key>::base_contains(Container&, PyObject*) [with Container = std::vector<bool, std::allocator<bool> >; DerivedPolicies = boost::python::detail::final_vector_derived_policies<std::vector<bool, std::allocator<bool> >, true>; bool NoProxy = true; bool NoSlice = false; Data = bool; Index = long unsigned int; Key = bool]':
> /usr/local/include/boost/python/suite/indexing/indexing_suite.hpp:284:30: note: 'x' declared here
>   284 |                 extract<Key> x(key);
>       |                              ^
> ```

GCC's `Wmaybe-uninitialized` is really flaky, and I've seen a bunch of times that it is outright wrong. It is one warning I would generally advocate disabling.

core/src/G3Timestream.cxx
arahlin: Since these functions are all static, they actually aren't being used anywhere.  I think people make explicit copies and use the in-place operators instead.  I suspect these implementations are holdovers from before we let numpy handle all the operator business on that end.

arahlin:
This warning popped up with gcc-14, and I'm also not sure of the right way to fix it -- naively removing the `&`'s in lines 101 and 102 caused test failures, presumably because `vec` is no longer a reference to the thing it's supposed to be a reference to.

```
In file included from /Users/arahlin/code/spt/spt3g_software_public/core/src/G3Vector.cxx:1:
/Users/arahlin/code/spt/spt3g_software_public/core/include/core/pybindings.h: In instantiation of 'int pyvector_getbuffer(PyObject*, Py_buffer*, int, const char*) [with T = double; PyObject = _object]':
/Users/arahlin/code/spt/spt3g_software_public/core/src/G3Vector.cxx:120:55:   required from here
  120 |         return pyvector_getbuffer<G3VectorDouble::value_type>(obj, view, flags,
      |                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~
  121 |             "d");
      |             ~~~~
/Users/arahlin/code/spt/spt3g_software_public/core/include/core/pybindings.h:101:31: error: possibly dangling reference to a temporary [-Werror=dangling-reference]
  101 |         const std::vector<T> &vec =
      |                               ^~~
/Users/arahlin/code/spt/spt3g_software_public/core/include/core/pybindings.h:102:68: note: the temporary was destroyed at the end of the full expression 'boost::python::extract<const std::vector<double>&>(selfobj).boost::python::extract<const std::vector<double>&>::boost::python::converter::extract_reference<const std::vector<double>&>.boost::python::converter::extract_reference<const std::vector<double>&>::operator()()'
  102 |             boost::python::extract<const std::vector<T> &>(selfobj)();
      |             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
/Users/arahlin/code/spt/spt3g_software_public/core/include/core/pybindings.h: In instantiation of 'int pyvector_getbuffer(PyObject*, Py_buffer*, int, const char*) [with T = long long int; PyObject = _object]':
/Users/arahlin/code/spt/spt3g_software_public/core/src/G3Vector.cxx:127:52:   required from here
  127 |         return pyvector_getbuffer<G3VectorInt::value_type>(obj, view, flags,
      |                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~
  128 |             "q");
      |             ~~~~
/Users/arahlin/code/spt/spt3g_software_public/core/include/core/pybindings.h:101:31: error: possibly dangling reference to a temporary [-Werror=dangling-reference]
  101 |         const std::vector<T> &vec =
      |                               ^~~
/Users/arahlin/code/spt/spt3g_software_public/core/include/core/pybindings.h:102:68: note: the temporary was destroyed at the end of the full expression 'boost::python::extract<const std::vector<long long int, std::allocator<long long int> >&>(selfobj).boost::python::extract<const std::vector<long long int, std::allocator<long long int> >&>::boost::python::converter::extract_reference<const std::vector<long long int, std::allocator<long long int> >&>.boost::python::converter::extract_reference<const std::vector<long long int, std::allocator<long long int> >&>::operator()()'
  102 |             boost::python::extract<const std::vector<T> &>(selfobj)();
      |             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
```

core/include/core/G3NetworkSender.h
arahlin: It's a user argument, so I figured that a negative value might mean something specific, but maybe that's not the case.

CMakeLists.txt
arahlin: I agree.  @nwhitehorn any objection to not using `-Werror`?

core/include/core/G3NetworkSender.h
cnweaver: It would technically do something (never allow anything 'low priority' to be put in the queue), but it seems doubtful that anyone wanted to do that.

arahlin:
Aha!  Figured it out.  Both of these warnings are caused by `boost::python::extract` variables being created and immediately evaluated without checking if they're valid.  I believe this is now fixed in 20ae15d

CMakeLists.txt
nwhitehorn: I agree there are some downsides, but I normally like -Werror since it forces people to fix questionable code. I am willing to be over-ruled, of course, but it seems like it came in handy in a few places here.

core/src/G3Timestream.cxx
nwhitehorn: I believe so, yes.

nwhitehorn:
> Aha! Figured it out. Both of these warnings are caused by `boost::python::extract` variables being created and immediately evaluated without checking if they're valid. I believe this is now fixed in [20ae15d](https://github.com/CMB-S4/spt3g_software/commit/20ae15d985b1bb0a03224c343e9f60c7f8e47906)

Are these always necessary? It is a lot of boilerplate to do the check() and I would imagine it would just fine without it so long as the extract isn't a temporary or the code doesn't extract a reference.

arahlin:
> > Aha! Figured it out. Both of these warnings are caused by `boost::python::extract` variables being created and immediately evaluated without checking if they're valid. I believe this is now fixed in [20ae15d](https://github.com/CMB-S4/spt3g_software/commit/20ae15d985b1bb0a03224c343e9f60c7f8e47906)
> 
> Are these always necessary? It is a lot of boilerplate to do the check() and I would imagine it would just fine without it so long as the extract isn't a temporary or the code doesn't extract a reference.

All of the places where I made this change seem to be necessary (at least in terms of clearing up these warnings).  I believe there are more places where I haven't made this change that I left as they were.  Seems like this is necessary when (a) dealing with extracting pointer objects or (b) calling `.check()` without then doing the extraction.  The latter results in a `maybe uninitialized` error that happens on deletion of the `bp::extract` object.

core/src/G3Timestream.cxx
arahlin: I've added these to the header so that they're available in C++, but let numpy handle the python side.

arahlin:
> GCC's `Wmaybe-uninitialized` is really flaky, and I've seen a bunch of times that it is outright wrong. It is one warning I would generally advocate disabling.

In general I would agree, but I'm hesitant here, because this particular warning seems to pop up only in gcc-14 and only for the two sets of boolean vector bindings (`std::vector<bool>` and `G3VectorBool`).  We are handling `bool`'s slightly differently from the other scalar types, so maybe there's something missing in the implementation?

CMakeLists.txt
arahlin: What if we disable `-Werror` by default, but enable it specifically in the github workflow, so that warnings are quickly caught there?

arahlin:
Ok, I've been convinced.  `std::vector<bool>` is weird.  I've added some `pragma GCC diagnostic` blocks to ignore the warnings on the `vector_indexing_suite.hpp` header altogether.

CMakeLists.txt
nwhitehorn: I'd prefer to keep it on by default, but I am willing to be outvoted.

core/include/core/serialization.h
```diff
@@ -65,14 +65,14 @@ struct g3frameobject_picklesuite : boost::python::pickle_suite
 };
 
 template <class T>
-inline int
+inline uint32_t
```
nwhitehorn, line 68: Why uint32_t specifically, given the cast below, rather than a generic unsigned?

core/src/G3Frame.cxx
```diff
@@ -149,7 +150,8 @@ static std::string FrameObjectClassName(G3FrameObjectConstPtr obj)
 		}
 	}
 
-	return cxx_demangle(typeid(*obj).name());
+	auto &r = *obj;
+	return cxx_demangle(typeid(r).name());
```
nwhitehorn, line 154: Why does this help? I don't see an obvious reason to warn about the original code...

core/src/G3Pipeline.cxx
```diff
@@ -31,8 +31,10 @@ cxx_demangle(const char *name)
 void
 G3Pipeline::Add(G3ModulePtr module, std::string name)
 {
-	if (name == "")
-		name = cxx_demangle(typeid(*module).name());
+	if (name == "") {
+		auto& r = *module;
+		name = cxx_demangle(typeid(r).name());
```
nwhitehorn, line 36: Same question here: I don't understand the original compiler warning in this case.

maps/src/G3SkyMapMask.cxx
```diff
@@ -639,9 +639,16 @@ PYBINDINGS("maps")
 	    "Apply a mask in-place to the mask, optionally inverting which "
 	    "pixels are zeroed.  If inverse = False, this is equivalent to "
 	    "in-place element-wise logical-and with the mask.")
+#ifdef __clang__
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wself-assign-overloaded"
+#endif
```
nwhitehorn, line 645: I would lean in favor of globally disabling this warning, which seems dubious and will be guaranteed to trigger on certain idioms in Boost.

core/src/G3Frame.cxx
arahlin: The warning was:

```
/Users/arahlin/code/spt/spt3g_software_public/core/src/G3Frame.cxx:152:29: warning: expression with side effects will be evaluated despite being used as an operand to 'typeid' [-Wpotentially-evaluated-expression]
        return cxx_demangle(typeid(*obj).name());
                                   ^
```

core/include/core/serialization.h
arahlin: ¯\_(ツ)_/¯ why not?

maps/src/G3SkyMapMask.cxx
arahlin: Fair.  Does it make sense to add to the pybindings.h header as a global pragma for any source files that use that header?

CMakeLists.txt
arahlin: I agree that we should have it enabled *somewhere* so that we can catch and fix warnings, but I'm with Chris that the penalty is too high for a user that just wants the package to work and hasn't done anything other than upgrade their compiler.  I've make sure that the configuration in the github workflow is sufficient to catch this.

core/include/core/serialization.h
cnweaver: Because cereal [documents](https://uscilab.github.io/cereal/assets/doxygen/structcereal_1_1detail_1_1Version.html) it as being `uint32_t`? We're not likely to meet a compiler which implements `unsigned` as 16 bit, but I think it is allowed.

maps/src/G3SkyMapMask.cxx
arahlin: I've done the above, rather than opening the door again to creating a laundry list of particular warnings to be disabled in the cmake file.

core/src/G3Frame.cxx
arahlin: I've pushed a different solution that templates `cxx_demangle` instead, based on one of the comments in [this](https://stackoverflow.com/questions/46494928/clang-warning-on-expression-side-effects) post.

maps/src/G3SkyMapMask.cxx
nwhitehorn: I'd honestly prefer it in CMakeLists.txt to a semi-hidden global #pragma

core/src/G3Frame.cxx
nwhitehorn: Weird, OK. Thanks!

maps/src/G3SkyMapMask.cxx
nwhitehorn: (After conversation with Sasha offline, this seems fine)

nwhitehorn (approved):

PR #161, opened 2024-09-16
Use smart cmake targets for linking

arahlin:
Make sure that library linking targets have correctly attached include directories, so that any other targets linked against them automatically use those include directories as well.

Use the `REQUIRED` keyword with `find_package()` to make sure that required system libraries are found.

Update FLAC and NetCDF find modules to newer versions with smart targets.

Bump minimum cmake version to 3.12.


arahlin:
It might be ok to remove all of the special handling for python and boost now that RHEL7 has reached EOL.  Thoughts?

nwhitehorn (approved):
Thanks! Looks good.

PR #162, opened 2024-09-20
Clean up core library python bindings

arahlin:
Distribute the monster python.cxx file out to the places where the relevant structures are defined, maintaining the registration order of various dependencies where necessary.

This PR is just splitting off these portions of #159 to both simplify that PR and ensure that these changes are not the source of the linking issue there (they are not).


nwhitehorn (approved):
Looks great, thanks!

PR #163, opened 2024-09-26
Separate shared libraries and python modules

arahlin:
Each of the sub-libraries (core, calibration, dfmux, gcp, maps) are separated into two parts:

A shared library called `libspt3g-<name>`, which is installed to `$CMAKE_INSTALL_PREFIX/lib` and contains all of the C++ structures defined in that library.  A shared module called `_lib<name>`, which is added as an extension module to the `spt3g` python package, and whose namespace is internally imported into the `spt3g.<name>` python module.  This allows linking the shared libraries against each other, while treating the modules as independent endpoints that just expose the appropriate components to python, thus removing the need for the `dload` and `load_pybindings` infrastructure.

Libraries link against the `Python::Module` cmake target and use dynamic lookup to avoid linking against `libpython` until their associated module is loaded by the interpeter.  Compiled executables then link against `libpython` explicitly via the `Python::Python` cmake target.

This PR is the last step toward producing a portable package that can be installed using standard python tools like `pip`.


arahlin:
This PR is a subset of #159, which has been approved, so I'll merge the two in stages.

PR #164, opened 2024-10-11
Consolidate IO stream handling in one place

arahlin:
This PR migrates all of the code that handles file and buffer IO stream creation into one place.  This ensures consistent handling of all input and output data serialization streams, and minimizes the footprint of the Boost iostreams and filesystem libraries in the core code, making it incrementally easier to eventually replace Boost altogether.

### DIFF
--- a/core/include/core/G3Reader.h
+++ b/core/include/core/G3Reader.h
@@ -3,7 +3,7 @@
 
 #include <string>
 #include <vector>
-#include <boost/iostreams/filtering_stream.hpp>
+#include <dataio.h>
 
 #include <G3Module.h>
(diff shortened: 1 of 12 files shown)


nwhitehorn (approved):

cnweaver (approved):
It would be nice if the new (and newly renamed) functions had doxygen-style documentation of parameters and return values (e.g. it is not obvious what `g3_istream_from_path()` returns without reading the implementation, and I'm not sure that even after doing that that I understand the point, and one can make the guess that timeout is measured in seconds, but putting it in black and white is always useful).

PR #166, opened 2024-10-14
Native quaternion implementation

### LINKED ISSUE #143: Incorrect behavior of `numpy.abs()` on `G3VectorQuat` objects
```
In [1]: from spt3g.core import quat, G3VectorQuat
   ...: import numpy as np
   ...: q = quat(1.0, 0.0, 0.0, 0.0)
   ...: v = G3VectorQuat([q, q, q, q])

In [2]: abs(v)  # the builtin function has the correct behavior
Out[2]: spt3g.core.G3VectorDouble([1, 1, 1, 1])

In [3]: np.abs(v)  # expect this to return a 1-d numpy array
Out[3]:
array([[1., 0., 0., 0.],
       [1., 0., 0., 0.],
       [1., 0., 0., 0.],
       [1., 0., 0., 0.]])
```

arahlin:
This PR removes the dependency on boost::math::quaternion in favor of a minimal native implementation, which includes simple arithmetic and comparison operators, and conj, abs, norm and pow function implementations.  The Quat object is serializeable and provides a numpy buffer interface.

Closes #143.


nwhitehorn:
I had one large-scale question. Making G3Quat a frame object instead of a primitive type makes a few things more complicated:
- It will slow down conversions between numpy arrays and quaternions and G3VectorQuat
- It will likely make a G3VectorQuat occupy a significant amount more memory
- Because of the less-dense format, it may impede some compiler vectorization operations

Have you looked at the change in storage size or other performance metrics?

arahlin:
I haven't yet.  That said, it would be straightforward to walk back the frameobject change and leave the serialization and vectorization as is.

maps/src/G3SkyMap.cxx
```diff
@@ -1536,7 +1536,7 @@ PYBINDINGS("maps") {
 	       "a disc of the given radius at the given sky coordinates.")
 
 	    .def("query_disc",
-	      (std::vector<size_t> (G3SkyMap::*)(quat, double) const)
+	      (std::vector<size_t> (G3SkyMap::*)(const Quat &, double) const)
```
nwhitehorn, line 1539: Is there any reason to do this? If quat is the same type as Quat basically, I'm not sure this is any more efficient.

arahlin: Fair, this made some more sense when Quat was derived from G3FrameObject.

arahlin: It looks like it's O(10ns) of overhead for the value argument vs const reference.  Pretty small, but might add up?

nwhitehorn:
This basically looks good to me. It looks like it would reduce the diff by quite a lot if you did s/Quat/quat. Is there a reason to change the class name?

arahlin:
> This basically looks good to me. It looks like it would reduce the diff by quite a lot if you did s/Quat/quat. Is there a reason to change the class name?

Not a strong reason, just to differentiate this class from the boost one, and to make it clear that it's a class object.  But functionally it should be the same as the boost variant, so maybe it's ok to leave it as `quat`?

core/src/G3Quat.cxx
```diff
@@ -430,10 +691,55 @@ pow(const G3TimestreamQuat &a, int b)
...
+	if (view == NULL) {
+		PyErr_SetString(PyExc_ValueError, "NULL view");
+		return -1;
+	}
+
+	view->shape = NULL;
+
+	bp::handle<> self(bp::borrowed(obj));
+	bp::object selfobj(self);
+	bp::extract<QuatPtr> ext(selfobj);
```
nwhitehorn, line 714: Why use a shared pointer here instead of a reference (e.g. `Quat &`)? Using a shared pointer adds a bunch of overhead to no obvious purpose and requires defining the types QuatPtr etc.

nwhitehorn (approved):

PR #167, opened 2024-10-18
First pass at removing Boost library dependencies

arahlin:
This PR replaces various Boost utilities with standard library equivalents where possible.  The changes here are all the pieces that have straightforward replacements.  Remaining uses outside of this PR:

* `boost::math::quaternion`: Replaced with a native G3FrameObject implementation in a separate PR (see #166)
* `boost::filesystem`: Can be easily replaced with `std::filesystem`, but requires C++17.
* `boost::iostreams`:  Intertwined with serialization and python bindings, but the implementation is now entirely contained in core/dataio.cxx/h (see #164)
* `boost::python`:  Yes.

### DIFF
--- a/cmake/G3Test.h
+++ b/cmake/G3Test.h
@@ -10,6 +10,9 @@
 #include <string>
 #include <vector>
 
+#define STRINGIZE(s) STRINGIZE_DIRECT(s)
+#define STRINGIZE_DIRECT(s) #s
+
 namespace G3Test{
 
 class TestFailure{
@@ -88,7 +91,7 @@ inline void testEquivalence(const std::string& file, unsigned int line,
 /// TEST_GROUP(MyTests)
 #define TEST_GROUP(GROUPNAME) \
 namespace{ \
-static const char* GetTestGroup(){ return #GROUPNAME; } \
+static const char* GetTestGroup(){ return STRINGIZE(GROUPNAME); } \
 }
 
 /// Define a test. Use like:
@@ -100,7 +103,7 @@ static const char* GetTestGroup(){ return #GROUPNAME; } \
 static void test_implementation_ ## TESTNAME(); \
 namespace{ \
 static G3Test::TestRegisterer register_ ## TESTNAME (GetTestGroup(), \
-	#TESTNAME, \
+	STRINGIZE(TESTNAME), \
 	test_implementation_ ## TESTNAME); \
 } \
 static void test_implementation_ ## TESTNAME()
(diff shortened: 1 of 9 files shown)


cmake/G3Test.h
```diff
@@ -89,7 +88,7 @@ inline void testEquivalence(const std::string& file, unsigned int line,
 /// TEST_GROUP(MyTests)
 #define TEST_GROUP(GROUPNAME) \
 namespace{ \
-static const char* GetTestGroup(){ return BOOST_PP_STRINGIZE(GROUPNAME); } \
+static const char* GetTestGroup(){ return #GROUPNAME; } \
```
cnweaver, line 91: Direct use of the stringizing operator is not as general as `BOOST_PP_STRINGIZE`, as it does not allow other macros to be expanded. It can be replaced in two lines, though:

```
#define STRINGIZE(s) STRINGIZE_DIRECT(s)
#define STRINGIZE_DIRECT(s) #s
```

core/include/core/pybindings.h
```diff
@@ -174,7 +173,10 @@ class G3ModuleRegistrator {
 //     SPT3G_PYTHON_MODULE(foo)
 // for a package whose name will be _libfoo
 #define SPT3G_PYTHON_MODULE(name) \
-BOOST_PYTHON_MODULE(BOOST_PP_CAT(_lib, name))
+BOOST_PYTHON_MODULE(_lib ## name)
```
cnweaver, line 176: Same issue as with `BOOST_PP_STRINGIZE`, but likely less concerning, as probably no one wants to generate a module name from a macro.

core/src/dataio.cxx
```diff
@@ -24,9 +23,9 @@ int
 g3_istream_from_path(g3_istream &stream, const std::string &path, float timeout)
 {
 	stream.reset();
-	if (boost::algorithm::ends_with(path, ".gz"))
+	if (!path.compare(path.size() - 3, 3, ".gz"))
```
cnweaver, line 26: This will throw if the path is too short for the substring to exist.

cnweaver (changes requested):
Only some small issues.

core/include/core/pybindings.h
arahlin: Agreed, I'm not super worried about this piece, so I'll leave it as is here.  If we start making more complicated module names, I'm sure this whole `SPT3G_PYTHON_MODULE()` macro will change all over again.

cnweaver (approved):

PR #168, opened 2024-11-18
Replace boost::shared_ptr with std::shared_ptr

arahlin:
Boost version 1.63 and newer [supports](https://github.com/boostorg/python/issues/29) `std::shared_ptr` python converters.  Thus, in this PR, all boost shared pointers are replaced with standard library shared pointers everywhere in the library.  The required C++ standard is also increased to C++17, since shared pointers for allocated C arrays do not work with the older standards.

In the process, this PR also updates the cereal library to the latest version (1.3.2) without any patching.


nwhitehorn (approved):
Thanks for this long-overdue set of changes! I think this is ready to merge. It is probably worth noting in the commit message when you do that you have updated cereal as well.

cnweaver (approved):
I see no issues.

tskisner:
Thanks for this work!  I confirm that (with a couple small modifications), the downstream `so3g` package builds against this branch.  Looks good to me.

tskisner (approved):
Looks great!

PR #169, opened 2024-11-20
Replace boost::filesystem with std:filesystem

arahlin:
With the update to the C++17 standard in #168, we can also drop the `boost::filesystem` dependency.


nwhitehorn (approved):
Yes, please!

PR #172, opened 2025-01-23
Better G3Reader.tell() method

arahlin:
Implement a byte counter as a boost istream filter to avoid buffering in the kernel.  This should speed up read IO for frame indexing.

### DIFF
--- a/core/include/core/G3Reader.h
+++ b/core/include/core/G3Reader.h
@@ -11,10 +11,10 @@ class G3Reader : public G3Module {
 public:
 	G3Reader(std::string filename, int n_frames_to_read = -1,
                  float timeout = -1., bool track_filename = false,
-	         size_t buffersize = 1024*1024);
+	         size_t buffersize = 1024*1024, bool counter = false);
 	G3Reader(std::vector<std::string> filenames, int n_frames_to_read = -1,
                  float timeout = -1., bool track_filename = false,
-	         size_t buffersize = 1024*1024);
+	         size_t buffersize = 1024*1024, bool counter = false);
 
 	void Process(G3FramePtr frame, std::deque<G3FramePtr> &out);
 	off_t Seek(off_t offset);
@@ -32,6 +32,7 @@ private:
 	float timeout_;
 	bool track_filename_;
 	size_t buffersize_;
+	bool counter_;
 
 	SET_LOGGER("G3Reader");
 };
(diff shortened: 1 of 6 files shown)


core/src/G3Reader.cxx
```diff
@@ -100,14 +102,17 @@ void G3Reader::Process(G3FramePtr frame, std::deque<G3FramePtr> &out)
...
-		return g3_istream_seek(stream_, offset);
-	} catch (...) {
-		log_fatal("Cannot seek %s; stream closed at EOF.", cur_file_.c_str());
-	}
+	return g3_istream_seek(stream_, offset);
+	// try {
+	// 	return g3_istream_seek(stream_, offset);
+	// } catch (...) {
+	// 	log_fatal("Cannot seek %s; stream closed at EOF.", cur_file_.c_str());
+	// }
```
nwhitehorn, line 110: Presumably this being commented out isn't intentional? Can the code just be deleted?

core/include/core/G3Reader.h
```diff
@@ -11,10 +11,10 @@ class G3Reader : public G3Module {
 public:
 	G3Reader(std::string filename, int n_frames_to_read = -1,
                  float timeout = -1., bool track_filename = false,
-	         size_t buffersize = 1024*1024);
+	         size_t buffersize = 1024*1024, bool counter = false);
 	G3Reader(std::vector<std::string> filenames, int n_frames_to_read = -1,
                  float timeout = -1., bool track_filename = false,
-	         size_t buffersize = 1024*1024);
+	         size_t buffersize = 1024*1024, bool counter = false);
```
nwhitehorn, line 17: Does counter = true cause a measurable performance drop? Or can we just make it also be on?

core/src/G3Reader.cxx
arahlin: Ah, that was for debugging.  I've replaced it with a similar error to `tell()` that requires `counter=True`.

core/include/core/G3Reader.h
arahlin: I haven't tried.  Maybe @mhasself or @tskisner can test this out for their use case?

arahlin: Testing actually indicates that this counter introduces a measurable performance drop, so we may need a different approach here.

PR #173, opened 2025-02-13
Use sys/time.h on all systems for gettimeofday()

tskisner:
On some linux systems, `gettimeofday()` is only declared in `sys/time.h` (not `time.h`).  This PR just includes `sys/time.h` on all systems.  It seems like [that is the standard header to include](https://pubs.opengroup.org/onlinepubs/009604599/functions/gettimeofday.html).


arahlin (approved):
Thanks!

PR #175, opened 2025-02-19
Updates to wheel building

tskisner:
- Bump vendored version of boost to 1.87, to enable support for numpy-2.

- Bump cibuildwheel version and add a dependabot workflow to keep actions up to date.

- Add a fix for macos wheel builds (see https://github.com/pypa/cibuildwheel/issues/2021)

- Add python-3.13 to the wheel build matrix.

- Change no-as-needed linking option to be based on OS flavor rather than compiler.

- Compile libflac as well as boost when building wheels.  Bump this vendored version to 1.5.0 which includes threaded compression.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -87,7 +87,7 @@ jobs:
         env:
           CIBW_BUILD: ${{ matrix.build }}
           CIBW_BEFORE_ALL_LINUX: yum install -y bzip2-devel netcdf-devel
-          CIBW_BEFORE_ALL_MACOS: echo "" > /dev/null
+          CIBW_BEFORE_ALL_MACOS: brew install netcdf
           CIBW_BEFORE_BUILD_MACOS: >
             ln -s $(dirname $(readlink -f $(which python3)))/python3-config $(dirname $(which python3))/python3-config &&
             pip install cmake && ./deps/install_flac.sh && ./deps/install_boost.sh
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -5,6 +5,10 @@ on:
   release:
     types: [ published ]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build_wheels:
     name: Build wheels for ${{ matrix.build }}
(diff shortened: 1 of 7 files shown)


tskisner:
Just a note that I [triggered the wheel checks here](https://github.com/tskisner/spt3g_software/actions/runs/13406690335).  I removed the netcdf support on the macos wheels to avoid bringing in homebrew dependencies, but I can try to re-enable that and see if anything breaks.

.github/workflows/wheels.yml
```diff
@@ -71,17 +83,33 @@ jobs:
...
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2.20.0
+        uses: pypa/cibuildwheel@v2.22.0
         env:
           CIBW_BUILD: ${{ matrix.build }}
-          CIBW_BEFORE_ALL_LINUX: yum install -y bzip2-devel flac-devel netcdf-devel
-          CIBW_BEFORE_ALL_MACOS: brew install bzip2 flac netcdf
-          CIBW_BEFORE_BUILD: ./deps/install_boost.sh
+          CIBW_BEFORE_ALL_LINUX: yum install -y bzip2-devel netcdf-devel
+          CIBW_BEFORE_ALL_MACOS: echo "" > /dev/null
```
arahlin, line 90: I would try to build with netcdf if possible.  There are a good number of Mac users that would use these wheels for streaming data from readout boards to KST or similar.

arahlin (changes requested):

.github/workflows/wheels.yml
tskisner: Sounds good, [wheel build attempt here](https://github.com/tskisner/spt3g_software/actions/runs/13420593679).  I didn't know what netcdf was used for.  Note that netcdf depends on hdf5, which for homebrew packages is built with gcc-14.  If this causes problems, we could try to build our whole stack with homebrew gcc-14 rather than building all with clang as we are currently doing.  Let's see what happens.

tskisner: Ok, linking to homebrew netcdf seemed to work in the build (not sure if any of the unit tests exercise those features).  Note that I excluded the homebrew bzip2 package since that is "keg only" and not placed into `/usr/local/lib` (to avoid conflicts with MacOS provided bzip2).  So we have always been linking to the system bzip2 and would need to pass extra flags to cmake if we wanted to use the homebrew one.

arahlin: I think the only tests in the dfmux library are to make sure that compiled components import correctly.  Could add a test to make sure netcdfdump actually works, but I think if the import test is passing and netcdf compiles and links, then we're ok.

arahlin (approved):
Looks great, thanks for sorting this out!

Should remove comments about netcdf from the PR summary, otherwise this looks like it's ready to merge.

PR #177, opened 2025-03-04
Link against stdc++fs library for compatibility with older compilers

arahlin:
This linking should in principle work with all versions of g++ and clang that support this library.


arahlin:
Fixes SouthPoleTelescope/spt3g_software#353

hivon:
Works fine under Linux for gcc 8.5 to 14.2

PR #178, opened 2025-03-09
Use STL streams for buffer IO

arahlin:
This reduces the `boost::iostreams` footprint to only file IO.


nwhitehorn (approved):
This looks good to me. Is performance the same? Some of the existing code was to work around subtle but important performance issues in either Boost or the STL.

arahlin:
I've tested reading and writing frames to pickle files, and don't see any appreciable difference in performance between old and new code.  I did make sure to keep your fixes for `xsgetn` in place for the read buffer implementation, and am using it more broadly than the G3Frame.load method.  Aside from that, it seems the boost and std output buffer implementations are largely the same.

PR #179, opened 2025-03-10
Add support for LZMA compression of G3 files

arahlin:


nwhitehorn (approved):
This is great, thank you!

PR #180, opened 2025-03-10
Improve library linking

arahlin:
* Use private links for libraries that are specific to one module
* Separate linking of Boost::python and Boost::iostreams
* Remove explicit links to private libraries in exported cmake config


arahlin:
@tskisner not sure if this PR will make your life easier or not, but thought I'd tag you as an FYI.

nwhitehorn (approved):

Issue #125, opened 2023-10-18
Frame-wise compression


arahlin:
Useful for allowing fseek on compressed frames

nwhitehorn:
Is there a reason we _want_ to use fseek? The way we encode pipeline state, and so the whole processing model, kind of break if we do that. For one-off debugging, this is somewhat useful, but you still need some kind of index of where you should fseek to, which means you need to read the whole file once anyway, so just decompressing it seems easy enough...

arahlin:
The so3g tools use this functionality.  We added indexing with fseek/ ftell  to the G3Reader already in a previous PR.  This open issue is just closing the loop, so that this functionality works on compressed data as well.

PR #181, opened 2025-03-12
Consistent G3Frame versioning and serialization

arahlin:
This PR uses the same versioning and archive serialization methods for G3Frame instances as those for all other serializable objects in the library.

The G3Frame class version is defined in the header, along with specialized load/save methods for IO with cereal binary archives.

The load/save methods for streams are renamed to loads/saves to ensure that the cereal library binds the correct serialization methods.

This PR does not change the serialized blob structure or increment the serialization version, so full compatibility with existing data is maintained.

### DIFF
--- a/core/include/core/G3Frame.h
+++ b/core/include/core/G3Frame.h
@@ -96,8 +96,12 @@ public:
 	G3Frame &operator = (const G3Frame &);
 
 	// Serialize (or deserialize) from an IO stream.
-	template <typename T> void save(T &os) const;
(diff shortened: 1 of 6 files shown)


arahlin:
The intention here is to facilitate implementing #125 or similar features.

nwhitehorn (approved):
This works because of a lucky coincidence in how Cereal stores version codes, but, given that lucky coincidence, I don't see any downside here.

PR #182, opened 2025-03-13
Use STL seekg/tellg for G3Reader frame indexing

arahlin:
This fixes a regression caused by the boost::iostreams variant.

Profiling files with many thousands of frames reproduces the problem and verifies the solution.


arahlin:
This PR supersedes #172.

nwhitehorn (approved):

PR #183, opened 2025-03-18
RebinMaps pipeline module

arahlin:
Useful for consistent rebinning of all maps in a map frame.


arahlin:
Discussed and tested offline.

PR #187, opened 2025-03-19
Hide boost::iostreams implementation behind smart pointers

arahlin:
This allows eventually rewriting the file IO ABI while ensuring STL iostream compatibility.

### DIFF
--- a/core/include/core/G3Reader.h
+++ b/core/include/core/G3Reader.h
@@ -3,7 +3,7 @@
 
 #include <string>
 #include <vector>
-#include <dataio.h>
+#include <iostream>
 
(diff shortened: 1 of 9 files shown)


nwhitehorn:
Have you tested this for performance? I would expect potentially-significant regressions for ARC files, but the rest of it is probably fine.

arahlin:
> Have you tested this for performance? I would expect potentially-significant regressions for ARC files, but the rest of it is probably fine.

I see no difference in performance on my laptop when reading one 130MB arc file with a bunch of small frames.

nwhitehorn (approved):
Thanks for testing!

PR #189, opened 2025-03-22
Remove dependency on Boost.IOStreams

### LINKED ISSUE #185: Replace Boost.IOStreams with STL constructs

arahlin:
Implement file IO functions using STL constructs.

A set of stream buffer classes are defined in the private `streams.h` header.  API functions in the public `dataio.h` header are significantly reduced, relying on the stream buffers to provide the backend implementation for flush, seek, tell, etc.  The callback registration API is used to ensure destruction of instantiated stream buffers.

The compression implementation here is API-compatible with the boost variant, verified by writing files with this branch and reading files with the master branch, and vice versa.  These compression streambuf classes could straightforwardly be used in other applications, for example to implement the hybrid bzip2/flac compression scheme used in G3SuperTimestreams.

Seek/tell efficiency of the reader is maintained, although slightly reimplemented here with a caching counter.

New features:
- Both reading and writing streams use a (tunable) 1MB buffer, for compressed, uncompressed, local and remote streams.
- Update ARCFileReader API to expose the timeout and buffersize istream options.

Fixes #185.


arahlin:
Performance testing showed that the slow `tellg` issue cropped up again, so I implemented a byte counter that seems to work.  Otherwise read/write speeds are the same as the boost implementation.

nwhitehorn (approved):
This looks good to me from an architectural perspective. I'm not 100% I understand the rationale for the file-extension arguments, but that's a minor quibble. Thanks!

core/include/core/dataio.h
```diff
@@ -19,11 +20,11 @@
  *                  read until EOF.
  * @param  timeout  Timeout in seconds for socket connections.
  * @param  buffersize Advisory buffer size in bytes for aggregating reads
- * @return stream   The input stream configured by this function.
+ * @param  ext      Required file extension (excluding compression suffixes)
```
cnweaver, line 23: The `ext` parameters to `g3_*stream_*_path` don't seem to do much; they only provide a very weak check that files being opened might be what users expect them to be? They add a lot of complexity to do this, though, and the resulting error message ("Invalid filename") is rather uninformative. Given that this check was not present before, was there any significant rate of user mistakes that it would help?

core/src/streams.h
```diff
@@ -0,0 +1,352 @@
...
+		log_fatal("Seek not implemented for remote stream");
+	}
+
+	std::streampos seekpos(std::streampos pos, std::ios_base::openmode mode) {
+		log_fatal("Seek not implemented for remote stream");
+	}
+
+private:
+	int fd_;
+	std::vector<char> buffer_;
```
cnweaver, line 79: The various `*Buffer` classes never seem to resize their `buffer_`s or do any other high-level operations on them, so it seems like `std::unique_ptr<char[]>` would be a more natural type than `std::vector<char>`. (This would require explicitly keeping a variable for `size`, but this would still be a small savings compared to the `vector` keeping both a size and a capacity around.)

core/src/streams.h
```diff
@@ -0,0 +1,352 @@
...
+	}
+
+	std::streampos seekoff(std::streamoff off, std::ios_base::seekdir way,
+	    std::ios_base::openmode mode) {
+		if (!(mode & std::ios_base::in))
+			log_fatal("Seek not implemented for output stream");
+		// short-circuit for tellg
+		if (off == 0 && way == std::ios_base::cur)
+			return bytes_;
+		std::streampos n = file_.rdbuf()->pubseekoff(off, way, mode);;
```
cnweaver, line 112: Extra semicolons hre and in `seekpos`.

core/src/streams.h
```diff
@@ -0,0 +1,352 @@
...
+		log_fatal("Seek not implemented for remote stream");
+	}
+
+private:
+	int fd_;
+	std::vector<char> buffer_;
+	size_t bytes_;
+};
+
+class InputFileStreamCounter : public std::streambuf {
```
cnweaver, line 83: I still don't grasp what `InputFileStreamCounter` and `OutputFileStreamCounter` accomplish; they are just wrappers around `std::filebuf`s contained within extra `fstream` objects, but do some extra bookkeeping of the number of bytes read or written, which is redundant with, and sometimes replaced by, the tracking already done by the underlying `filebuf`?
Applying the following patch, everything works completely equivalently, with less code:
```
--- a/core/src/dataio.cxx
+++ b/core/src/dataio.cxx
@@ -260,7 +260,10 @@ g3_istream_from_path(std::istream &stream, const std::string &path, float timeou
 #endif
 	default:
 		// Read buffer
-		stream.rdbuf(new InputFileStreamCounter(path, buffersize));
+		std::unique_ptr<std::filebuf> buf(new std::filebuf);
+		if(!buf->open(path, std::ios::in | std::ios::binary))
+			log_fatal_stream("Error opening file " << path);
+		stream.rdbuf(buf.release());
 		break;
 	}
 
@@ -307,7 +310,13 @@ g3_ostream_to_path(std::ostream &stream, const std::string &path, bool append,
 		break;
 #endif
 	default:
-		stream.rdbuf(new OutputFileStreamCounter(path, buffersize, append));
+		std::unique_ptr<std::filebuf> buf(new std::filebuf);
+		std::ios_base::openmode mode = std::ios::out | std::ios::binary;
+		if (append)
+			mode |= std::ios::app;
+		if(!buf->open(path, mode))
+			log_fatal_stream("Error opening file " << path);
+		stream.rdbuf(buf.release());
 		break;
 	}
 ```

File: core/src/dataio.cxx [hunk 5]
```diff
@@ -127,136 +116,201 @@ connect_remote(const std::string &path, float timeout)
...
-		stream->push(boost::iostreams::counter64());
-	std::ios_base::openmode mode = std::ios::binary;
-	if (append)
-		mode |= std::ios::app;
-	stream->push(boost::iostreams::file_sink(path, mode));
+	if (fpath.has_parent_path()) {
+		auto ppath = fpath.parent_path();
+		if (!std::filesystem::exists(ppath))
+			log_fatal("Parent path does not exist: %s",
+			    ppath.string().c_str());
```
cnweaver, line 194: Is there any utility in `check_output_path` checking that the parent path is a directory?

cnweaver (changes requested):
As a general correctness issue, I don't see any calls to `ios::xalloc`, but it is a precondition that `ios::pword` can only be called with index values previously returned by `xalloc`. (I think the point here is that _any_ user code is allowed to use `pword` and `iword` to attach things to `ios_base`, so each distinct use should call `xalloc` to ensure that it is using a unique index to do so. Hard-coding it to zero everywhere is unsafe, because it can conflict with any other (rare) code using this feature. It looks like there should be central bit of code that calls `xalloc` once, either the first time the index is needed, or on startup, and caches the value it gets in a variable which every function calling `pword` should read to get the index to use.

core/src/streams.h
arahlin: The InputFileStreamCounter caches the byte count and short-circuits tellg (i.e. `seekoff(0, cur, in)`) by returning the cached value.  On my system, I see something like 10x system time used if don't cache the bytes read.

core/src/dataio.cxx
arahlin: IMO `has_parent_path()` returning true indicates that the parent has to be a directory, otherwise it wouldn't be a parent path.  I think that's sufficient here, but I could add a check for is_directory(ppath) if you think its necessary.

cnweaver: > has_parent_path() returning true indicates that the parent has to be a directory

The reason I suggest the extra check is that it doesn't. It indicates that the user _thinks_ it is a directory and wrote a path with that assumption, but that may not be the state of the filesystem:
```
$ cat write_fail.py 
#!/usr/bin/env python3
from spt3g import core
with core.G3Writer("test_dir/foo/bar.g3") as w:
	f = core.G3Frame()
	w(f)
$ ./write_fail.py 
FATAL (Unknown) 29-Mar-2025:08:29:02 EDT: Error opening file test_dir/foo/bar.g3 (dataio.cxx:320 in void g3_ostream_to_path(. . . ))
Traceback (most recent call last):
  File "./write_fail.py", line 3, in <module>
    with core.G3Writer("test_dir/foo/bar.g3") as w:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Error opening file test_dir/foo/bar.g3 (in void g3_ostream_to_path(. . . ))
# The parent path exists:
$ ls test_dir/
foo
# But it is not a directory!
$ file test_dir/foo
test_dir/foo: ASCII text
```

core/include/core/dataio.h
cnweaver: `Filename 'somefile' does not have expected extension 'blah'` would at least help the user know why they're being stopped and maybe what to do about it.

core/src/streams.h
cnweaver: For completeness, copying the rest of the discussion from Slack:
Part of my concern is that on one of my systems (Mac OS 10.15 on x86_64), the proposed caching method causes a factor of four increase in user time when writing. I can't reproduce the increase in system time without the cache (maybe it affects only Arm Macs? Only Mac OS >=16?), but I'm working on an approach which will hopefully avoid both issues by using a subclass of filebuf directly.

cnweaver: That commit does fix the regression I was seeing, and is equivalent to the prototype I was working on.

core/src/dataio.cxx
arahlin: Ah, good point.  Yep,  I'll add that check!

core/src/streams.h
arahlin: Great! I did still end up having to count bytes separately, because whatever filebuf does under the hood for tellg seems to balloon the system time a whole bunch.  Deriving from filebuf directly does make a bunch more sense though, thanks for catching that.

cnweaver (approved):

PR #190, opened 2025-03-26
Add a Simons Observatory sub-module.

tskisner:
This PR is just a proof of principle / reference implementation showing an up-streaming of the `G3SuperTimestream` class from Simons Observatory tools:

- To ease downstream use and avoid the need for bundling spt3g core into the `so3g` package, create a sub-module called "so".

- Add the only unique class used by S.O. data (G3SuperTimestream) to the `so` submodule, along with some numpy helper tools.

- Modify the integrated G3SuperTimestream source slightly to allow compilation with -Wall -Werror on both gcc and clang.

This work may be superceded by a shim that translates to/from `G3TimestreamMap` on the fly.  TBD.

### DIFF
--- a/so/CMakeLists.txt
+++ b/so/CMakeLists.txt
@@ -1,0 +1,26 @@
+# Find the numpy include path
+execute_process(COMMAND ${Python_EXECUTABLE} -c
+  "import numpy; print(numpy.get_include())"
+  OUTPUT_VARIABLE NUMPY_INCLUDE_DIR
+  OUTPUT_STRIP_TRAILING_WHITESPACE
+)
+
+add_spt3g_library(so SHARED
+  src/G3SuperTimestream.cxx
+  src/numpy_assist.cxx
+)
+
+target_include_directories(so PUBLIC ${NUMPY_INCLUDE_DIR})
+target_link_libraries(so PUBLIC core)
+
+find_package(OpenMP)
+if(OpenMP_FOUND)
+  target_link_libraries(so PUBLIC OpenMP::OpenMP_CXX)
+  set(OpenMP_FOUND ${OpenMP_FOUND} PARENT_SCOPE)
+endif()
+
+add_spt3g_module(so src/python.cxx)
+
+link_python_dir()
+
+add_spt3g_test(imports)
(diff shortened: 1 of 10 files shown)


so/include/so/G3SuperTimestream.h
```diff
@@ -0,0 +1,123 @@
...
+// and not the base class' load/save.
+namespace cereal {
+    template <class A> struct specialize<
+        A, G3SuperTimestream, cereal::specialization::member_load_save> {};
+}
+
+G3_POINTERS(G3SuperTimestream);
+G3_SERIALIZABLE(G3SuperTimestream, 0);
+
+class g3supertimestream_exception : std::exception
```
arahlin, line 112: use the G3Logging stuff instead of creating a custom exception -- `SET_LOGGER("G3SuperTimestream")` in the class definition, and use `log_fatal()` instead of `throw g3supertimestream_exception`

so/include/so/exceptions.h
```diff
@@ -0,0 +1,128 @@
+#pragma once
```
arahlin, line 1: Drop this header file in favor of standard python exceptions and G3Logging

so/tests/imports.py
```diff
@@ -0,0 +1,2 @@
+import spt3g._libso
+from spt3g._libso import G3SuperTimestream
```
arahlin, line 2: Add this test script: https://github.com/simonsobs/so3g/blob/master/test/test_g3super.py

so/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
+# Find the numpy include path
```
arahlin, line 1: Make this module build conditionally on finding Bzip2 and FLAC, since these are technically optional libraries in core.

arahlin (changes requested):
Some preliminary comments on this with some straightforward changes that would make it easier to merge.  We strictly avoid linking against numpy everywhere in the core library, and instead rely directly on the python buffer protocol for array access.  I think it's feasible to do the same here, but would take some work.  Linking against numpy headers is tricky because of the numpy 2 compatibility issues etc.  I'd rather avoid that altogether.

Separately, If the intention is to migrate / integrate this class with G3TimestreamMap, I think it would make sense to strip out any unused features.

arahlin:
Closing this in favor of #192

PR #192, opened 2025-04-01
Shim class for loading G3SuperTimestream objects from disk

arahlin:
This PR is the first step in integrating the G3SuperTimestream class used by the Simons Observatory team into the core library.  This class is a very thin wrapper around the core G3TimestreamMap class.  It is designed only to load G3SuperTimestream objects from G3 files on disk for immediate conversion to G3TimestreamMap objects.  It is also not exposed to the public or python API.


arahlin:
A couple of missing features that I've noticed so far:

- [ ] There's no check that the times array is contiguous and monotonic.  We should probably have some code that fills in the missing samples with nans.
- [ ] There's also no check that the names array is ordered, which G3TimestreamMap requires.  Probably the right thing to do is to populate the map in an ordered way.

core/CMakeLists.txt
```diff
@@ -5,6 +5,13 @@ else(APPLE)
 	set(CORE_EXTRA_SRCS "")
 endif(APPLE)
 
+if(NOT DEFINED WITH_SUPERTIMESTREAM)
+	set(WITH_SUPERTIMESTREAM TRUE CACHE BOOL "Build G3SuperTimestream class")
+endif()
+if(WITH_SUPERTIMESTREAM)
+	set(CORE_EXTRA_SRCS ${CORE_EXTRA_SRCS} src/G3SuperTimestream.cxx)
+endif()
+
```
nwhitehorn, line 14: Any reason to make this conditional?

core/src/G3Frame.cxx
```diff
@@ -358,6 +361,14 @@ void G3Frame::blob_decode(struct blob_container &blob)
...
 	cereal::PortableBinaryInputArchive item_ar(is);
 	item_ar >> make_nvp("val", ptr);
+
+#ifdef HAS_SUPERTIMESTREAM
+	// Convert to public type
+	auto v = std::dynamic_pointer_cast<G3SuperTimestream>(ptr);
+	if (!!v)
+		ptr = std::make_shared<G3TimestreamMap>(*v);
+#endif
+
```
nwhitehorn, line 371: This is slightly ugly, but I don't see any other real way to do it. I might make the comment a little longer to explain to some future reader why any of this exists.

core/CMakeLists.txt
arahlin: Largely for ease of transition on the SO side, I think.

core/src/G3Frame.cxx
arahlin: I figured out a less-ugly solution by specializing cereal's `PolymorphicCasters::upcast()` function.  Seems to do the trick nicely.

core/CMakeLists.txt
nwhitehorn: Maybe we can invert the sense of the flag then? Have a "don't build this" flag instead of a "build this" one?

core/include/core/G3Timestream.h
```diff
@@ -27,19 +27,21 @@ class G3Timestream : public G3FrameObject {
...
+	    units(None), use_flac_(0), flac_depth_(24),
+	    buffer_(new std::vector<double>(l, r)),
 	    data_(&(*buffer_)[0]), len_(buffer_->size()), data_type_(TS_DOUBLE) {}
 	virtual ~G3Timestream() {
 		if (buffer_) delete buffer_;
 	}
 
 	// FLAC compression levels range from 0-9. 0 means do not use FLAC.
 	void SetFLACCompression(int compression_level);
+	void SetBitDepth(int bit_depth);
```
nwhitehorn, line 44: Is there a reason to have this? If the version of FLAC supports 32-bit data, can't we just use 32-bit data? Otherwise, stay on 24-bit data?

core/src/G3SuperTimestream.cxx
```diff
@@ -0,0 +1,458 @@
...
+	template <class A> struct specialize<A, G3SuperTimestream,
+	    cereal::specialization::member_load_save> {};
+
+	// Convert to G3TimestreamMap before handing off to the user
+	namespace detail {
+		template <> inline std::shared_ptr<void>
+		PolymorphicCasters::upcast(std::shared_ptr<G3SuperTimestream> const & dptr,
+		    std::type_info const & baseInfo) {
+			return PolymorphicCasters::upcast(
+			    std::make_shared<G3TimestreamMap>(*dptr), baseInfo);
```
nwhitehorn, line 38: This is very cute.

arahlin: Is it just? :)

core/CMakeLists.txt
arahlin: Done -- `WITHOUT_SUPERTIMESTREAM`

tskisner (approved):
Thanks for this work!  For the record, I tested this branch in combination with a [draft PR of so3g](https://github.com/simonsobs/so3g/pull/204) that allows building against newer spt3g versions.  With that combination, I was able to load some Simons Observatory data through this shim into the downstream `sotodlib` and `toast` data containers.

arahlin:
I think this is ready to merge, and we can iterate on any missing features in G3TimestreamMap separately. Any objections, @nwhitehorn or @mhasself ?

PR #193, opened 2025-04-03
Preserve insertion order in G3TimestreamMap

arahlin:
Use a custom OrderedMap class that behaves identically to std::map, except that items are iterated over in the order they were inserted into the map.

This allows preserving the detector ordering when creating a G3TimestreamMap from a 2D array and a set of keys, e.g. when converting between G3SuperTimestream and G3TimestreamMap objects, as the former are not guaranteed to be ordered alphabetically.

Performance is significantly faster than the standard std::map implementation, and maintains the same G3TimestreamMap serialization version.

### DIFF
--- a/core/include/core/G3Timestream.h
+++ b/core/include/core/G3Timestream.h
@@ -272,7 +272,8 @@ public:
 		}
 	}
 
-	template <class A> void serialize(A &ar, unsigned v);
+	template <class A> void load(A &ar, unsigned v);
+	template <class A> void save(A &ar, unsigned v) const;
 	std::string Description() const;
 };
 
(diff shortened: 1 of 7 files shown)


nwhitehorn:
What effect does this have on look-up times? Removing the sort order on the keys seems like it could cause significant performance regressions when the number of detectors is large.

core/include/core/G3Timestream.h
```diff
@@ -5,7 +5,7 @@
 #include <G3TimeStamp.h>
 #include <G3Vector.h>
 #include <vector>
-#include <map>
+#include <containers.h>
```
nwhitehorn, line 8: Is there a way to avoid including containers.h here? It might be nice to hide this as an internal detail. I'm not sure it is possible though.

nwhitehorn (approved):

PR #194, opened 2025-04-03
Add support for full-depth compression of int32 timestreams

arahlin:
Add a bit_depth attribute to G3Timestream objects, allowing the FLAC algorithm to enable compression of the full 32 bits per sample, rather than restricting to 24 bits.  FLAC version 1.4 and newer supports 32-bit compression, so this is now the default setting; backward compatibility is maintained for the one use case relying on 24-bit compression: SPT-3G (dfmux) data written on site.

If 32-bit compression is selected, then int64 and double timestreams are truncated to int32 prior to compression, then decompressed into their original data type.


nwhitehorn:
Is there a reason to make this a configurable parameter instead of unconditionally doing 32-bit compression on newer FLAC versions?

arahlin:
> Is there a reason to make this a configurable parameter instead of unconditionally doing 32-bit compression on newer FLAC versions?

I was mainly thinking in terms of backward compatibility.  It looks like clustertools py3-v4 uses FLAC version 1.3.3, so will have to still keep the 24-bit code around, as I'm not sure the decompression will do the right thing on data compressed with newer FLAC?  If there's no issue with that, then I'd be fine with just setting the encoder bit depth conditionally on FLAC version.

core/include/core/G3Timestream.h
```diff
@@ -27,19 +27,21 @@ class G3Timestream : public G3FrameObject {
...
+	    units(None), use_flac_(0), flac_depth_(24),
+	    buffer_(new std::vector<double>(l, r)),
 	    data_(&(*buffer_)[0]), len_(buffer_->size()), data_type_(TS_DOUBLE) {}
 	virtual ~G3Timestream() {
 		if (buffer_) delete buffer_;
 	}
 
 	// FLAC compression levels range from 0-9. 0 means do not use FLAC.
 	void SetFLACCompression(int compression_level);
+	void SetFLACDepth(int bit_depth);
```
cnweaver, line 44: If I'm reading correctly, `SetFLACDepth` and `GetBitDepth` are counterparts, so it would help if their names matched.

core/src/G3Timestream.cxx
```diff
@@ -122,9 +122,18 @@ template <class A> void G3Timestream::save(A &ar, unsigned v) const
...
+		if (data_type_ == TS_INT64 || data_type_ == TS_INT32)
+			data_type_out = TS_INT32;
+
+		ar & cereal::make_nvp("flac_depth", flac_depth_);
+		ar & cereal::make_nvp("data_type", data_type_out);
+
 		// Copy to 24-bit integers
 		inbuf.resize(size());
-		switch (data_type_) {
+	        switch (flac_depth_) {
```
cnweaver, line 134: Looks like this line is accidentally indented with spaces.

core/src/G3Timestream.cxx
```diff
@@ -149,6 +158,31 @@ template <class A> void G3Timestream::save(A &ar, unsigned v) const
 				break;
 			default:
 				log_fatal("Unknown timestream datatype %d", data_type_);
+			}
+			break;
+		case 32:
+			switch (data_type_) {
+			case TS_DOUBLE:
+				for (size_t i = 0; i < size(); i++)
+					inbuf[i] = int32_t(float(((double *)data_)[i]));
```
cnweaver, line 167: Is the cast to `float` intentional? It seems like it will casue the data to be reduced to 24-bit precision, causing values in the range (2^24, 2^32) to lose their lower 8 bits unnecessarily.

cnweaver (changes requested):
I think this makes good sense in general, just a few nitpicks.

core/include/core/G3Timestream.h
arahlin: My instinct was to match the SetFLACCompression/GetCompressionLevel pair, but not a strong preference.  I'll change that.

core/src/G3Timestream.cxx
arahlin: Yeah, I found some weird behavior when testing otherwise, where the decompressed data didn't match what I would have expect by simply casting double -> int32.  Maybe I'm misunderstanding something about what flac is doing under the hood.

arahlin: Ok, this is now sorted, preserving the data type through compression / decompression was the way to go.

nwhitehorn:
> > Is there a reason to make this a configurable parameter instead of unconditionally doing 32-bit compression on newer FLAC versions?
> 
> I was mainly thinking in terms of backward compatibility. It looks like clustertools py3-v4 uses FLAC version 1.3.3, so will have to still keep the 24-bit code around, as I'm not sure the decompression will do the right thing on data compressed with newer FLAC? If there's no issue with that, then I'd be fine with just setting the encoder bit depth conditionally on FLAC version.
> 
> EDIT: From the documentation it does seem like the older version can't decompress 32-bit data. I think it makes sense to leave the option as a configurable choice, for groups that exclusively use a sufficiently recent version of FLAC. The alternative is that we upgrade clustertools py3-v4 to use FLAC 1.4.3, but I assume that's not an option.

My understanding was that the reference FLAC decoder has been able to decode 32-bit data since forever -- it's part of the spec -- but the encoder could not. I would pretty strongly prefer to avoid more manual knobs here.

nwhitehorn:
Looking at the code, the only real thing that happens if you set flac_depth to 24 vs. 32 is that it clears the high-order 8 bits for you if you set 24. I'm not sure that saves us from anything -- does the library fail if they are set or does it just ignore them? If not, can't we just let it silently do the right thing?

arahlin:
I just tested this using clustertools py3-v4, and setting the newer encoder to use 32 bits per sample causes an older decoder to throw an `unparseable stream` error when trying to load the newer data, so I think we're stuck with this implementation.

core/src/G3Timestream.cxx
```diff
@@ -182,7 +224,7 @@ template <class A> void G3Timestream::save(A &ar, unsigned v) const
 		FLAC__StreamEncoder *encoder = FLAC__stream_encoder_new();
 		FLAC__stream_encoder_set_channels(encoder, 1);
 		// XXX: should assert if high-order 8 bits are not clear
```
nwhitehorn, line 226: I think this comment is now obsolete.

core/src/G3Timestream.cxx
```diff
@@ -232,6 +274,34 @@ template <class A> void G3Timestream::save(A &ar, unsigned v) const
 #endif
 }
 
+template <typename T>
+std::vector<T> *
+unpack_flac(const std::vector<int32_t> &buf, uint8_t nanflag, const std::vector<bool> &nanbuf)
+{
+	// Represent data as floats internally if possible. These have the
+	// same significand depth (24 bits) as the max. bit depth of the
+	// reference FLAC encoder we use, so no data are lost, and allow
```
nwhitehorn, line 283: As is this one.

nwhitehorn (approved):
Two small nits about obsolete comments in the code; approved otherwise.

PR #196, opened 2025-04-17
Fix ang_to_quat missing return statement

### LINKED ISSUE #195: Missing return statement in maps.quathelpers.ang_to_quat
In [`maps/python/quathelpers.py`](https://github.com/CMB-S4/spt3g_software/blob/master/maps/python/quathelpers.py):
```
@usefulfunc
def ang_to_quat(alpha, delta, start=None, stop=None):
    """
    Convert a set of angles (or vector of them) specified as a (longitude,
    latitude) pair to a pointing quaternion (or vector of them). If start
    and stop are defined, the return value for vectors will be a
    G3TimestreamQuat with start and stop times set to the provided values.
    """

   ... # Omitted code

    if len(q) == 1:
        return Quat(q[0, :])
    else:
        if start is not None:
            out = G3TimestreamQuat(q)
            out.start = start
            out.stop = stop
        else:
            return G3VectorQuat(q)
```

`ang_to_quat` seems to be missing a return statement for the case in which start and stop times are provided. The function returns `None` instead of the `G3TimestreamQuat`, `out`, as the documentation suggests.

jonahjlee:
Resolves #195.

- Fix `ang_to_quat` by adding the missing return statement.
- Add a simple test for `ang_to_quat` using the `start` and `stop` arguments.

This is my first contribution so please let me know if any changes are needed.

### DIFF
--- a/maps/python/quathelpers.py
+++ b/maps/python/quathelpers.py
@@ -60,6 +60,7 @@ def ang_to_quat(alpha, delta, start=None, stop=None):
             out = G3TimestreamQuat(q)
             out.start = start
             out.stop = stop
+            return out
         else:
             return G3VectorQuat(q)
 
--- a/maps/tests/quatangtest.py
+++ b/maps/tests/quatangtest.py
@@ -20,3 +20,8 @@ assert(np.allclose(maps.ang_to_quat(alpha, delta), core.G3VectorQuat([maps.c_ang
 
 assert(np.allclose(maps.quat_to_ang(maps.c_ang_to_quat_(angle[0], angle[1])), angle))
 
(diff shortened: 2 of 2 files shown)


arahlin (approved):
Thanks for the catch!

arahlin:
(Feel free to squash-and-merge at your leisure!)

jonahjlee:
Thanks! I don't think that I have the required permissions to merge though, so you might have to do it for me.

arahlin:
> Thanks! I don't think that I have the required permissions to merge though, so you might have to do it for me.

Oh, right!  Can do.

PR #200, opened 2025-05-09
Documentation Overhaul

arahlin:
Refactor documentation tools to make use of various extensions, including:

* sphinx.ext.napoleon for flexible docstring parsing
* sphinx.ext.autodoc for auto-documentation
* sphinx-rtd-theme for a more full-featured web interface

The docstring parser in spt3g.core now simply generates a lean RST file with autodoc directives for each object that is to be added to the API documentation. The sphinx conf.py file includes Documenter subclasses that parse the various
C++ binding docstrings so that overloaded functions and type annotations are displayed as true Python.

Additionally, the output moddoc.rst file has been expanded to include many of the objects that have thus far lacked documentation, e.g. non-serialized class types.

A build of this documentation is accessible at https://sptlocal.grid.uchicago.edu/~arahlin/docs/index.html

Enums and namespaces (G3Units, G3Constants) are not documented here, but will likely be more straightforward to handle with pybind11.


nwhitehorn (approved):
This looks great! One nit: is there a reason to separate the "class-like" modules and "function-like" modules? They work the same way for the user, so separating them seems a bit odd.

arahlin:
> This looks great! One nit: is there a reason to separate the "class-like" modules and "function-like" modules? They work the same way for the user, so separating them seems a bit odd.

The class-like ones can often be created outside of a pipeline and cache things for later (e.g. `maps.CoaddMaps`).  They sometimes also have extra methods attached to them (e.g. `core.G3Reader`).

nwhitehorn:
Fair enough. Maybe you can set something about that difference to the intro text?

nwhitehorn:
Thanks! This looks great.

PR #201, opened 2025-05-15
Fix gzip compression

arahlin:
The GZip compression implementation used the base ZLib stream functions and did not account for the necessary GZip header and footer.  This PR reimplements the GZipDecoder and GZipEncoder using the correct gzread/gzwrite and associated functions.

This code has been verified to be able to read gzip-compressed files written by the original Boost.iostreams implementation.

### DIFF
--- a/core/src/compression.cxx
+++ b/core/src/compression.cxx
@@ -1,59 +1,5 @@
 #include "streams.h"
 
-#ifdef ZLIB_FOUND
-GZipDecoder::GZipDecoder(const std::string& path, size_t size) : Decoder(path, size)
-{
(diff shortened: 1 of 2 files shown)


arahlin:
I believe when I first wrote this I tested writing gzip files with the new code and reading with the old code, which seemed to work.  I overlooked testing the inverse, and apparently that was broken.  I'm not entirely sure how, as the windowBits setting is supposed to account for the gzip header correctly, but refactoring seems to have resolved the problem.

arahlin:
Superceded by d448ed3

PR #225, opened 2026-06-23
Allow slicing FlatSkyMap objects along one dimension

arahlin:
Slicing into a FlatSkyMap object along both dimensions returns a new map object with the same coordinate system.  This PR also enables slicing along only one of the two dimensions, in which case the returned map object has length 1 along the other dimension.  This behavior previously raised an error.


tcrawfor (approved):
Seems to work as advertised, thanks.

PR #226, opened 2026-06-30
Add bounding_box and crop methods to FlatSkyMap class

axf295:
Add this bounding box method to flatskymap objects.


maps/src/FlatSkyMap.cxx
```diff
@@ -561,6 +561,36 @@ FlatSkyMap::NonZeroPixels(std::vector<uint64_t> &indices,
 	}
 }
 
+std::vector<int>
```
arahlin: Should use `ssize_t` everywhere here instead of `int`, mainly for consistency.

arahlin (changes requested):
Looks good, one nit.

maps/src/FlatSkyMap.cxx
```diff
@@ -561,6 +561,36 @@ FlatSkyMap::NonZeroPixels(std::vector<uint64_t> &indices,
...
+FlatSkyMap::BoundingBox(double pad) const
```
arahlin: Perhaps also `pad` should be in pixels instead of angular units?  That seems more intuitive to me.

arahlin (changes requested):

maps/src/FlatSkyMap.cxx
```diff
@@ -561,6 +561,54 @@ FlatSkyMap::NonZeroPixels(std::vector<uint64_t> &indices,
 	}
 }
 
+std::vector<ssize_t>
+FlatSkyMap::BoundingBox(ssize_t x_pad, ssize_t y_pad) const
```
arahlin:
```suggestion
std::vector<size_t>
FlatSkyMap::BoundingBox(size_t x_pad, size_t y_pad) const
```

maps/src/FlatSkyMap.cxx
```diff
@@ -561,6 +561,54 @@ FlatSkyMap::NonZeroPixels(std::vector<uint64_t> &indices,
...
+	}
+
+	if (x_max < 0)
+		return {};
+
+	return {
+		std::max(y_min - y_pad, (ssize_t)0),
+		std::min(y_max + y_pad, (ssize_t)ypix_ - 1),
+		std::max(x_min - x_pad, (ssize_t)0),
+		std::min(x_max + x_pad, (ssize_t)xpix_ - 1)
```
arahlin:
```suggestion
		(ssize_t)std::max(y_min - y_pad, 0),
		(ssize_t)std::min(y_max + y_pad, ypix_ - 1),
		(ssize_t)std::max(x_min - x_pad, 0),
		(ssize_t)std::min(x_max + x_pad, xpix_ - 1)
```

arahlin (changes requested):

axf295 (commented):
I see what you meant w the crop_to now. yes that is nice. 
 
not sure what a lot of the cpp details you've added mean but user-side it looks good!

arahlin (approved):

